PR #3, opened 2020-06-30
Cm branch

cynthia-m:


code/src/main/webapp/CSS/style.css
```diff
@@ -6,6 +6,8 @@ body, html {
 
 h1, p, a {
   margin: 0;
+  color:#3293d3;
```
nvergel: Can we make the text a lot darker. Right now its a little hard to read. Maybe have it be an offset of black like #222

code/src/main/webapp/CSS/style.css
```diff
@@ -71,54 +77,72 @@ div {
...
+  }
+} 
+
+.landing-background #blue-buttons-to-procede {
   top: 60%;
 }
 
 #blue-buttons-to-procede button {
-  background: #116dff;
+  background: #ffffff;
```
nvergel: Not sure how I feel about the white buttons. Even thought the background is "light" everything else should not be. The white buttons don't pop out.

code/src/main/webapp/JavaScript/script.js
```diff
@@ -47,16 +47,46 @@ class PageController {
...
         this.currentlyShown = this.loginPage;
+
+        var listOfClassesHiddenStatus = {"landing" : true, "login" : false};
+        this.setHiddenStatus(listOfClassesHiddenStatus);
+
         break;
     }
   }
+
+  setHiddenStatus(listOfClassesHiddenStatus) {
```
nvergel: This method hides and unhides respective pages? If so can we get rid of hideCurrentPage()

code/src/main/webapp/JavaScript/script.js
```diff
@@ -47,16 +47,46 @@ class PageController {
...
+        this.setHiddenStatus(listOfClassesHiddenStatus);
+
         break;
       case "login":
         history.pushState({page: pageToShow}, 'Login', pageToShow);
         this.loginPage.show();
         this.currentlyShown = this.loginPage;
+
+        var listOfClassesHiddenStatus = {"landing" : true, "login" : false};
+        this.setHiddenStatus(listOfClassesHiddenStatus);
```
nvergel: There is a lot of repeated code in the two cases, partially my fault for no making a function that handles the logic for both. Therefore can you make a function that will do everything inside the cases so that each case only has two lines: the helper function and break.

code/src/main/webapp/JavaScript/script.js
nvergel: On second thought I am not sure how I feel about this method. I think Grady wanted each page to have its own show and hide function. However this method does that all of that at once.

code/src/main/webapp/html/loginChoices.html
```diff
@@ -0,0 +1,41 @@
...
+  <script>
+    const PAGE_CONTROLLER = new PageController();
+    PAGE_CONTROLLER.show('landing');
+    window.onpopstate = function(event) {
+      PAGE_CONTROLLER.show(event.state.page);
+    }
+  </script>
+  
+</body>
+</html>
```
nvergel: Is this a necessary file?

cynthia-m:
@googlebot I signed it!

nvergel (changes requested):
I left specific comments. Here are two broader comments:

.gitignore file at root should make it so that code/target is not added. However, for those files to be ignored they first have to be removed. This will look something like `git rm -r code/target/*`. Please do that. Additionally there is a file in the html folder that might not have to be there so you should get in the habit of only adding what is necessary (and maybe deleting old files like the former backgrounds).

For my second comment: the text disappears instantly without transitioning out. I think it should fade away with the background to make it look more cohesive. Also, can the title go from saying Welcome to saying Log-In or something like that? And maybe the explanation can go from "We are here to help..." to "In order to gather more data to help with contact tracing, you can choose to sign in through Google or ... ". Anyways, these are a general changes to the UI that, as a group, we can discuss later on.

code/src/main/webapp/CSS/style.css
cynthia-m: The background actually has kind of a medium tone so neither dark nor light contributes to contrast, so I'll add a border

code/src/main/webapp/JavaScript/script.js
cynthia-m: I'm fine with either, I made this method specifically for the single page but we can do either.

code/src/main/webapp/html/loginChoices.html
cynthia-m: Nope, deleted.

code/src/main/webapp/JavaScript/script.js
cynthia-m: done

code/src/main/webapp/CSS/style.css
cynthia-m: done

code/src/main/webapp/CSS/style.css
cynthia-m: done, made buttons dark with white text

code/src/main/webapp/CSS/style.css
```diff
@@ -1,124 +1,307 @@
...
+}
+.fade-out {
+  animation: fadeOut ease 1s;
+  -webkit-animation: fadeOut ease 1s;
+  -moz-animation: fadeOut ease 1s;
+  -o-animation: fadeOut ease 1s;
+  -ms-animation: fadeOut ease 1s;
+}
+
+.FAQ {
```
gradywardgoogle: meganit - css classes should alwasy be lower-dash-case

The reasons for this are actually pretty messed up - older internet browsers don't always support upper cased CSS Properties!

gradywardgoogle: using faq instead of FAQ throughout will be good!

code/src/main/webapp/CSS/style.css
```diff
@@ -1,124 +1,307 @@
...
+  padding: 2rem;
+  text-decoration: none;
+  top: 60%;
 }
 
-#line-horizontal-top {
-  top: 80%;
-  transition-delay: 1s
+#blue-buttons-to-procede button:hover {
+  box-shadow: 0 0 2rem #8bccf4;
```
gradywardgoogle: This is pretty subtle given the background changes. Perhaps make the box shaddow more of a transparent grey, so that it overlays nicely on the background regardless of the background color.

code/src/main/webapp/JavaScript/script.js
```diff
@@ -44,17 +44,67 @@ class PageController {
...
-        this.landingPage.show();
-        this.currentlyShown = this.landingPage;
+        var listOfClassesHiddenStatus = {"landing" : false, "login" : true};
+        this.setPageState("landing", this.landingPage, pageToShow, listOfClassesHiddenStatus)
         break;
       case "login":
-        history.pushState({page: pageToShow}, 'Login', pageToShow);
-        this.loginPage.show();
-        this.currentlyShown = this.loginPage;
+        var listOfClassesHiddenStatus = {"landing" : true, "login" : false};
```
gradywardgoogle: This will require updating every time we add a new page. Instead, let's do this.

Initialize this varaible above with every page set to true.
in each case statement, set one of the values to "false" (or even better, set the value of pageToShow to false 'listOfClassesHiddenStatus[pageToShow] = false').

code/src/main/webapp/JavaScript/script.js
```diff
@@ -44,17 +44,67 @@ class PageController {
...
+        this.setPageState("login", this.loginPage, pageToShow, listOfClassesHiddenStatus)
         break;
     }
   }
+
+  setPageState(whichPage, thisPage, pageToShow,  listOfClassesHiddenStatus){
+    if(whichPage === "landing") {
```
gradywardgoogle: nit: space before (, both here, line 57 and 60.

code/src/main/webapp/JavaScript/script.js
```diff
@@ -44,17 +44,67 @@ class PageController {
...
+          elements[index].classList.remove("fade-in");
+        } else {
+          elements[index].classList.remove("fade-out");
+          elements[index].classList.add("fade-in");
+        }
+      } 
+    }
+    setTimeout(this.setHiddenStatus2, 900, listOfClassesHiddenStatus);
+  }
+  setHiddenStatus2(listOfClassesHiddenStatus) {
```
gradywardgoogle: can we name this method a little clearer?

code/src/main/webapp/html/faq.html
```diff
@@ -0,0 +1,22 @@
...
+</head>
+<body>
+  <div class="general-background">
+    
+  </div>
+  <div class="FAQ" style="text-align:center, display: inline-block">
+      <div id="FAQ-q1" style="text-align:center, display: inline-block">
+        <h1 class="FAQ-questions"> What is Contact Tracing? </h1>
+        <p class="FAQ-answers"> Contact tracing is the process of identification of persons who may have come into contact with an infected person and subsequent collection of further information about these contacts. 
+          <br>TL;DR If you're positive for COVID-19, we help let people you've been in contact with know.<p>
```
gradywardgoogle: nit - avoid using the BR tag. Two ways of approaching this:

have faq-answers be a <div> that contains two <p>s, or have two <p>s each of which has an faq-answers class.

code/src/main/webapp/html/faq.html
```diff
@@ -0,0 +1,22 @@
...
+   <meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
+   <link rel="stylesheet" href="../CSS/style.css">
+   <script src="../JavaScript/script.js"></script>
+   <title></title>
+</head>
+<body>
+  <div class="general-background">
+    
+  </div>
+  <div class="FAQ" style="text-align:center, display: inline-block">
```
gradywardgoogle: Please move these to a css style.

gradywardgoogle (approved):
This is great work. Solid algorithms, much cleaner design, solid Object Oriented Programming. Big growth CL, good work!

PR #4, opened 2020-07-06
Setup objectify, data types and testing

nvergel:


code/src/main/java/com/google/sps/data/BusinessNumber.java
```diff
@@ -0,0 +1,33 @@
...
+  long phoneNumber;
+  private long getTimeWhenCalledUnixTimeSeconds;
+  private boolean businessHasBeenCalled = false;
+
+  public BusinessNumber(String nameOfBussiness, long phoneNumber) {
+    this.nameOfBussiness = nameOfBussiness;
+    this.phoneNumber = phoneNumber;
+  }
+
+  // Record time bussiness was contacted
```
Josiah-Caldwell: Feels like you should override setSuccess, getStatus, and getTime... even if they are empty in the interface.

code/src/main/java/com/google/sps/data/PersonEmail.java
```diff
@@ -0,0 +1,29 @@
...
+  String nameOfSender;
+  String email;
+  long getTimeWhenEmailedUnixTimeSeconds;
+  boolean personHasBeenEmailed;
+
+  public PersonEmail(String nameOfSender, String email) {
+    this.nameOfSender = nameOfSender;
+    this.email = email;
+  }
+
```
Josiah-Caldwell: Feels like you should override setSuccess, getStatus, and getTime... even if they are empty in the interface.

code/src/main/java/com/google/sps/data/PositiveUser.java
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  public long getFirstLogin() {
+    return firstLogin;
+  }
+
+  public long getLastLogin() {
+    return lastLogin;
+  }
+
```
Josiah-Caldwell: I feel like updateEmailsSent should have an int param and go up by said value

code/src/main/java/com/google/sps/data/BusinessNumber.java
nvergel: This may be my lack of Java experience but I thought we only had to override methods form a superclass?

code/src/main/java/com/google/sps/data/PositiveUser.java
nvergel: I was thinking that too but I also thought we would only want to increment when an email is successfully sent?

code/src/test/java/com/google/sps/data/BusinessTest.java
```diff
@@ -0,0 +1,19 @@
+package com.google.sps.data;
+
+import org.junit.Test;
+import org.junit.Assert;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+ 
+@RunWith(JUnit4.class)
+public final class BusinessTest {
+
```
Josiah-Caldwell: I see that this is testing that the contact method works but what I would do is have two seperate tests, for one assert false and do not call contact and for the other call contact and assert true. This would avoid any hiccups from having multiple assertions in one test

Josiah-Caldwell: Could call the first one createBusinessNoContact and the second createContactedBusiness

code/src/test/java/com/google/sps/data/ContactStatusTest.java
```diff
@@ -0,0 +1,27 @@
+package com.google.sps.data;
+
+import org.junit.Test;
+import org.junit.Assert;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+ 
+@RunWith(JUnit4.class)
+public final class ContactStatusTest {
+
```
Josiah-Caldwell: Same as for business test, I would split both tests into two seperate methods each

code/src/test/java/com/google/sps/data/PositiveUserTest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.junit.Test;
+import org.junit.Assert;
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+import java.time.Instant;
+ 
+@RunWith(JUnit4.class)
+public final class PositiveUserTest {
+
```
Josiah-Caldwell: Same as with BusinessTest and ContacStatusTest, I'd split them up

code/src/test/java/com/google/sps/data/BusinessTest.java
Josiah-Caldwell: I may be wrong tho, I'd ask Ben and Grady about testing

Josiah-Caldwell (commented):
Overall good, I'd ask Ben and Grady about testing though.

code/src/test/java/com/google/sps/data/BusinessTest.java
nvergel: I like the idea, makes it easier to debug, will implement rn

code/src/main/java/com/google/sps/data/BusinessNumber.java
cynthia-m: you have to override all methods in from an interface too I think

Josiah-Caldwell: Overriding is in general something that you want to do when there are multiple implementations of the same method so that would go for interfaces as well as superclasses

code/src/main/java/com/google/sps/data/PositiveUser.java
Josiah-Caldwell: That actually makes sense, works for me

code/src/main/java/com/google/sps/data/Business.java
```diff
@@ -0,0 +1,28 @@
...
+  private long timeOfLastContactUnixTimeSeconds;
+
+  private Business() {}
+  
+  public Business(String placeId) {
+    this.placeId = placeId;
+  }
+
+  // Update timeOfContact to current time
+  public void contact() {
```
cynthia-m: rename to be updateTimeOfContact() pls

code/src/main/java/com/google/sps/data/BusinessNumber.java
```diff
@@ -0,0 +1,33 @@
+package com.google.sps.data;
+
+import java.time.Instant;
+
+// Keep track of information needed to contact a business
+public class BusinessNumber implements ContactStatus{
+
+  String nameOfBussiness;
```
cynthia-m: Business

code/src/main/java/com/google/sps/data/BusinessNumber.java
nvergel: okay

code/src/main/java/com/google/sps/data/Business.java
```diff
@@ -0,0 +1,31 @@
+package com.google.sps.data;
+
+import com.googlecode.objectify.annotation.Entity;
+import com.googlecode.objectify.annotation.Id;
+import java.time.Instant;
+
+@Entity
+public class Business {// This class is solely to check if a business was contacted in the last week.
```
gradywardgoogle: Please use Javadoc (here and throughout) on public methods and classes.

Javadoc best practices: https://www.tutorialspoint.com/java/java_documentation.htm

gradywardgoogle: Most of this CL will be comments.

code/src/main/java/com/google/sps/data/Business.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Id String placeId;
+  private long timeOfLastContactUnixTimeSeconds;
```
gradywardgoogle: UnixTime here is redundant, since time is in the name, and seconds is at the end. Assuming the Unix Epoch is reasonable since all computer systems use it. 

timeOfLastContactSeconds is ideal here.

code/src/main/java/com/google/sps/data/BusinessNumber.java
```diff
@@ -0,0 +1,36 @@
+package com.google.sps.data;
```
gradywardgoogle: Let's name this package "storage", rather than "data".

code/src/main/java/com/google/sps/data/BusinessNumber.java
```diff
@@ -0,0 +1,36 @@
...
+
+import java.time.Instant;
+
+// Keep track of information needed to contact a business
+public class BusinessNumber implements ContactStatus{
+
+  String nameOfBusiness;
+  long phoneNumber;
```
gradywardgoogle: Let's use a string for phone number rather than a long. Strings seem less likely to have strange errors (what happens if there is a parenthesis in the text we get back, for example?)

code/src/main/java/com/google/sps/data/BusinessNumber.java
```diff
@@ -0,0 +1,36 @@
...
+  private long getTimeWhenCalledUnixTimeSeconds;
+  private boolean businessHasBeenCalled = false;
```
gradywardgoogle: Please use a constructor to initialize all fields.

code/src/main/java/com/google/sps/data/Business.java
```diff
@@ -0,0 +1,31 @@
...
+
+  private Business() {}
```
gradywardgoogle: If this will never be called, we should remove it.

code/src/main/java/com/google/sps/data/Location.java
```diff
@@ -0,0 +1,20 @@
+package com.google.sps.data;
+
+/* Store a location based on longitude/latitude coordinates and
+ * the time interval when the user was present
+ */
+public class Location {
+
+  long longitudeE7;
```
gradywardgoogle: Getters/Setters? Or public fields?

code/src/main/java/com/google/sps/data/Location.java
```diff
@@ -0,0 +1,20 @@
+package com.google.sps.data;
+
+/* Store a location based on longitude/latitude coordinates and
+ * the time interval when the user was present
```
gradywardgoogle: Javadoc requires /**, not /*

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
```diff
@@ -0,0 +1,23 @@
+package com.google.sps.data;
+
+import com.googlecode.objectify.annotation.Entity;
+import com.googlecode.objectify.annotation.Id;
+import java.time.Instant;
+import java.util.ArrayList;
+
+@Entity
+public class NegativeUserLocations {
```
gradywardgoogle: It now occurs to me that we might have someone use this feature multiple times. How would we need to adjust this class to do that/

code/src/main/java/com/google/sps/data/Business.java
gradywardgoogle: Just re-upping this one last time. Every public class, and most public methods should have a javadoc describing what it does/is. 

Also throughout this change, we should make fields private, and have getters/setters. Getters/setters typically don't require documentation, unless they have additional logic.

gradywardgoogle (changes requested):
Thanks for sending this! Biggest comment is the first one - to use Javadoc + set up getters/setters.

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
```diff
@@ -0,0 +1,23 @@
...
+
+  @Id String userId;
+  ArrayList<Location> locations;
+  long timeCreatedUnixTimeSeconds;
+
+  private NegativeUserLocations() {}
```
cynthia-m: I'm curious about having a private empty constructor? but if you have reasoning that's fine

code/src/main/java/com/google/sps/data/PositiveUser.java
```diff
@@ -0,0 +1,41 @@
...
+  private long lastLoginUnixTimeSeconds;
+
+  private PositiveUser() {}
+  
+  public PositiveUser(String userId) {
+    this.userId = userId;
+    firstLoginUnixTimeSeconds = lastLoginUnixTimeSeconds = Instant.now().getEpochSecond();
+  }
+
+  public void setLastLogin() {
```
cynthia-m: did grady say public methods need comments?

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
cynthia-m: ^^^throughout

code/src/main/java/com/google/sps/data/Business.java
nvergel: Does every variable need getters and setters? There already are getters and setters for some variables.

nvergel: I thought about doing Javadoc but wasn't sure if it goes before or after @Entity. Also should I be adding @author and so on?

code/src/main/java/com/google/sps/data/Business.java
nvergel: Okay I changed everything form UnixTimeSeconds to Seconds.

code/src/main/java/com/google/sps/data/BusinessNumber.java
nvergel: I changed everything from package com.google.sps.data to  package com.google.sps.storage. I also changed the file name to reflect this

code/src/main/java/com/google/sps/data/Business.java
nvergel: Any class that is loaded into Objectify needs a private constructor

code/src/main/java/com/google/sps/data/Location.java
nvergel: I was thinking public fields, however this discussion is covered by the first comment

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
nvergel: Not sure I understand

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
nvergel: I made it private whenever there is a variable with getters and setters, however it seems like Grady wants all variables to be private and have getters and setters?

code/src/main/java/com/google/sps/data/PositiveUser.java
nvergel: Not sure

nvergel: But yes I can add comments to every method if yall want

code/src/main/java/com/google/sps/data/BusinessNumber.java
nvergel: I am not sure what to initialize getTimeWhenCalledSeconds to

code/src/main/java/com/google/sps/data/BusinessNumber.java
nvergel: Okay I changed it. I think this is the only variable that uses a bad data type for what is being stored?

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
cynthia-m: yeah I think that's better practice

code/src/main/java/com/google/sps/data/Business.java
gradywardgoogle: No need to add author. The important part is doing the description of _what it is_ alongside a description of any _no obvious or nonintuitive things about it_. Describing things in english is far easier for a person who isn't familiar with the code or ideas behind it to grok.

Your goal with documentation is to make it so that a technically capable reader without knowledge of project could navigate it and make changes effectively. Public interfaces are a huge part of that goal, so each should have a description of what they do.

Yes - we should not have publicly visible fields, so we will need getters on each. We might not need setters on each if we expect them to be readonly.

code/src/main/java/com/google/sps/data/Business.java
gradywardgoogle: Thanks!

code/src/main/java/com/google/sps/data/BusinessNumber.java
gradywardgoogle: Yep the rest look reasonable.

code/src/main/java/com/google/sps/data/BusinessNumber.java
gradywardgoogle: perhaps that's a signal that it should be an optional long then! if it hasn't been called yet, we want to know that and have it cased differently than if it has been.

code/src/main/java/com/google/sps/data/Business.java
gradywardgoogle: Ah, TIL!

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
gradywardgoogle: If a user logs in as a "negative user" once, and then does so again a week later, will we update thier data? What if they log in after 2+ weeks? How does the negative-user-journey change if they opt to use the service multiple times?

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
gradywardgoogle: Since it wasn't obvious to Cynthia or I that we have the empty constructor for Objectify, let's make it explicit.

Can you add a comment above or next to each empty constructor describing why we have it (nescessary for objectify?)

code/src/main/java/com/google/sps/data/PositiveUser.java
gradywardgoogle: We don't need them on pure getters or setters, but yes, all other methods should have  javadoc.

code/src/main/java/com/google/sps/data/Business.java
nvergel: I added comments.

nvergel: I added comments and made everything private with getters.

code/src/main/java/com/google/sps/data/BusinessNumber.java
nvergel: okay

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
nvergel: I think this is an outdated comment since I changed the data type and  this one is no longer implemented

code/src/main/java/com/google/sps/data/NegativeUserLocations.java
nvergel: done

code/src/main/java/com/google/sps/storage/data-storage/Business.java
```diff
@@ -0,0 +1,35 @@
...
+@Entity
+public class Business {
+
+  @Id private String placeId;
+  private long timeOfLastContactSeconds;
+
+  // Objecify requires one constructor with no parameters
+  private Business() {}
+  
+  public Business(String placeId) {
```
gradywardgoogle: I could imagine placeId being ambiguous - maybe have a comment above the field declaration describing "this placeId refers to the PlaceId used by the Google Maps APIs (or something similar)

code/src/main/java/com/google/sps/storage/data-storage/Business.java
```diff
@@ -0,0 +1,35 @@
...
+    this.placeId = placeId;
+    
+    // Initialize time of last contacted to be over a week so contactedInLastWeek returns false
```
gradywardgoogle: Sorry I couldn't find this coversation on the previous thread - why not use an optional instead? This splits the logic around when we send messages into two parts - here and wherever we discuss that. If we modified it to be a month, we would need to update this code, in addition to the other code. That's subopmtimal, and I think an optional provides us with better description of the actual state.

code/src/main/java/com/google/sps/storage/data-storage/Business.java
```diff
@@ -0,0 +1,35 @@
...
+    timeOfLastContactSeconds = Instant.now().getEpochSecond() - Constants.oneWeekInSeconds;
+  }
+
+  // Update timeOfContact to current time
```
gradywardgoogle: Nice - both the method and the comment. Since this is documenting the interface itself - rather than an implementation comment - you'll want it to be in Javadoc, and ideally a full sentence.

code/src/main/java/com/google/sps/storage/data-storage/Business.java
```diff
@@ -0,0 +1,35 @@
...
+  public void updateTimeOfContact() {
+    timeOfLastContactSeconds = Instant.now().getEpochSecond();
+  }
+
+  //Businesses should only be contacted once a week
```
gradywardgoogle: Same here.

code/src/main/java/com/google/sps/storage/data-storage/Constants.java
```diff
@@ -0,0 +1,8 @@
+package com.google.sps.storage;
+
+/**
+* This class keeps track of constants we need.
+*/
+public class Constants {
+  final static long oneWeekInSeconds = 604800L;
```
gradywardgoogle: I love this class, I think it's a great idea. Let's do a few small things:

1) Let's rename this constant "contactingBusinessesCooldownPeriodInSeconds" or something similar. That way we could change it to be 2w without changing any other code!
2) final and static constants actually use a different variable naming case in Java: TYPICALLY_UPPER_SNAKE_CASE (I know it doesn't make a ton of sense).
3) After we change the constant to be named relative to what it _does_ rather than what it _is_ do some renaming in the business class to rename methods (ex: stillInCooldownPreiodFromLastContact, for example).

code/src/main/java/com/google/sps/storage/data-storage/Location.java
```diff
@@ -0,0 +1,37 @@
+package com.google.sps.storage;
+
+/**
+* Store a location based on longitude/latitude coordinates and
+* the time interval when the user was present
+*/
+public class Location {
+
+  private long longitudeE7;
```
gradywardgoogle: Similar to place Id, would you add a comment above the field describing what it is (this is the latitude/longitude multiplied by 10^7)?

code/src/main/java/com/google/sps/storage/data-storage/Location.java
```diff
@@ -0,0 +1,37 @@
...
+  private long intervalEndSeconds;
+
+  public Location(long lng, long lat, long intervalStart, long intervalEnd) {
+    longitudeE7 = lng;
+    latitudeE7 = lat;
+    intervalStartSeconds = intervalStart;
+    intervalEndSeconds = intervalEnd;
+  }
+
+  public long getLongitude() {
```
gradywardgoogle: nit: let's name these getLongitudeE7, since that's what they are (if I saw get longitude as a method I'd assume it would return a double between -180 and 180.)

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserLocation.java
```diff
@@ -0,0 +1,54 @@
...
+*/
+@Entity
+public class NegativeUserLocation {
+
+  private String  userId;
+  @Index private long longitudeE7;
+  @Index private long latitudeE7;
+  @Index private long intervalStartSeconds;
+  @Index private long intervalEndSeconds;
+  @Id private Long key;
```
gradywardgoogle: the key here isn't initialized? You could store a UUID as the key (google UUID to learn about it, but it's basically a guarenteed unique string).

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.ArrayList;
+
+/**
+* This class stores the information needed to identify a negative user
+* at a specific place.
+*/
+@Entity
+public class NegativeUserPlace {
+
+  private String userId;
```
gradywardgoogle: nit: here and throughout, can we order by:
Id first, indexes second, and other fields third?

code/src/main/java/com/google/sps/storage/data-storage/Place.java
```diff
@@ -0,0 +1,36 @@
+package com.google.sps.storage;
+
+/**
+* Store a place's name and id
```
gradywardgoogle: meganit - let's try to use full sentences throughout. I won't block on that, but definitely for the future, and if you could do a rough pass now that'd be great.

code/src/main/java/com/google/sps/storage/data-storage/Place.java
```diff
@@ -0,0 +1,36 @@
...
+*/
+public class Place {
```
gradywardgoogle: (Feel free to do this in a follow up Pr, but you could do it now if you wanted)

We'll want meaningful descriptions of each of these entities in a toString() method (so we can have simple logging + clear error reporting). That can be in a follow up PR.

code/src/main/java/com/google/sps/storage/data-storage/PositiveUser.java
```diff
@@ -0,0 +1,60 @@
...
+
+  public long getFirstLoginInSeconds() {
+    return firstLoginSeconds;
+  }
+
+  public long getLastLoginInSeconds() {
+    return lastLoginSeconds;
+  }
+
+  public void incrementEmailsSent() {
```
gradywardgoogle: This is nice! it's exposing a smaller interface, but it's the interface you want to expose. That's great.

code/src/main/java/com/google/sps/storage/data-storage/PositiveUser.java
```diff
@@ -0,0 +1,60 @@
...
+    emailsSent++;
+  }
+
+  public boolean emailLimitReached() {
```
gradywardgoogle: please store the value of 200 in your constants file (with a descriptive name)

gradywardgoogle: As it is this seems to actually be the opposite (userCanStillSendEmails) rather than (userCanNoLongerSendEmails = limitReached). Please name a little bit more precicely.

code/src/main/java/com/google/sps/storage/data-storage/PositiveUserContacts.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public ArrayList<PotentialContact> getContacts() {
+      return contacts;
+  }
+
+  public long getTimeCreatedSeconds() {
+      return timeCreatedSeconds;
+  }
+
+  //TODO
```
gradywardgoogle: is this intentional? Will this just be adding them to the contacts list?

code/src/main/java/com/google/sps/storage/data-storage/PositiveUserLocations.java
```diff
@@ -0,0 +1,40 @@
...
+  private PositiveUserLocations() {}
+
+  public PositiveUserLocations(String userId, ArrayList<Location> locations) {
+    this.userId = userId;
+    this.locations = locations;
+    timeCreatedSeconds = Instant.now().getEpochSecond();
+  }
+
+  public String getUserId() {
+      return userId;
```
gradywardgoogle: nit: inconsistent spacing.

code/src/main/java/com/google/sps/storage/messaging/BusinessNumber.java
```diff
@@ -0,0 +1,50 @@
...
+
+  public String getName() {
+    return nameOfBusiness;
+  }
+
+  public String getPhoneNumber() {
+    return phoneNumber;
+  }
+
+
```
gradywardgoogle: nit: extraneous whitespace

code/src/test/java/com/google/sps/storage/BusinessTest.java
```diff
@@ -0,0 +1,23 @@
...
+import org.junit.runner.RunWith;
+import org.junit.runners.JUnit4;
+ 
+@RunWith(JUnit4.class)
+public final class BusinessTest {
+
+  @Test
+  public void createBusiness() {
+    Business business = new Business("Test");
+    Assert.assertFalse(business.contactedInLastWeek());
```
gradywardgoogle: nit: we generally want tests broken into three sections by whitespace: Arrange, Act, Assert. If Act and Assert are on the same line (as they are here) that's OK, but you still want space for the arrange block. I know this sounds funny, but it's actually REALLY helpful once you get into larger tests.

gradywardgoogle: nit: you can have a static import of assertTrue and assertFalse. That tends to be the preferred strategy for such things:

import static org.junit.Assert.assertFalse;
import static org.junit.Assert.assertTrue;

code/src/test/java/com/google/sps/storage/BusinessTest.java
```diff
@@ -0,0 +1,23 @@
...
+  }
+
+  @Test
+  public void contactBusiness() {
+    Business business = new Business("Test");
```
gradywardgoogle: using the arrange, act, assert model, each of these lines should have a newline between them.

code/src/main/java/com/google/sps/storage/messaging/BusinessNumber.java
```diff
@@ -0,0 +1,50 @@
...
+  // Record time bussiness was contacted
+  @Override
+  public void setSuccess() {
+    getTimeWhenCalledSeconds = Optional.of(Instant.now().getEpochSecond());
+    businessHasBeenCalled = true;
+  }
+
+  @Override
+  public boolean getStatus() {
```
gradywardgoogle: more descriptive name please!

code/src/main/java/com/google/sps/storage/messaging/BusinessNumber.java
```diff
@@ -0,0 +1,50 @@
...
+    this.nameOfBusiness = nameOfBusiness;
+    this.phoneNumber = phoneNumber;
+    getTimeWhenCalledSeconds = Optional.empty();
+    businessHasBeenCalled = false;
+  }
+
+  // Record time bussiness was contacted
+  @Override
+  public void setSuccess() {
+    getTimeWhenCalledSeconds = Optional.of(Instant.now().getEpochSecond());
```
gradywardgoogle: Nice use of optional!

code/src/main/java/com/google/sps/storage/messaging/ContactStatus.java
```diff
@@ -0,0 +1,12 @@
+package com.google.sps.storage;
+
+import java.util.Optional;
+
+/**
+* Interface to help identify if a notification was successful.
+*/
+public interface ContactStatus {
+  public void setSuccess();
```
gradywardgoogle: each of the names in this interface could be a bit better. What about:

markContactedSuccessfully()
hasBeenContactedSuccessfully()
getTimeWhenContacted()

gradywardgoogle (commented):
This looks really good! I'd love for Ben to take a look over it before you submit but I don't see any blocking issues. Good work Nico.

code/src/main/java/com/google/sps/storage/data-storage/Business.java
nvergel: I can see how it is ambiguous since the class is called Business and we don't use a "businessId". I added the comment however it might be better to rename the class.

code/src/main/java/com/google/sps/storage/data-storage/Business.java
nvergel: done, I'll leave the comment up this time in case there's more variables that need optional

code/src/main/java/com/google/sps/storage/data-storage/Business.java
nvergel: done

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserLocation.java
nvergel: I think I left a comment on DD for this one, I'll wait to check that out

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
nvergel: Right now I would like if userId was the id but it won't be unique. I'll wait to make changes after I look at DD

code/src/main/java/com/google/sps/storage/data-storage/PositiveUserContacts.java
nvergel: Yeah I can see how it was confusing. I think contacts from calendar will be stored when an instance is initialized. I changed names to reflect this.

code/src/main/java/com/google/sps/storage/data-storage/PositiveUserLocations.java
nvergel: I hate that the default spacing in cloud shell if 4 spaces

code/src/main/java/com/google/sps/storage/data-storage/PositiveUser.java
nvergel: done

code/src/test/java/com/google/sps/storage/BusinessTest.java
nvergel: Why is static required for the import?

code/src/test/java/com/google/sps/storage/BusinessTest.java
nvergel: done

code/src/test/java/com/google/sps/storage/BusinessTest.java
nvergel: okay

code/src/main/java/com/google/sps/storage/data-storage/PositiveUser.java
nvergel: done

code/src/main/java/com/google/sps/storage/data-storage/Place.java
nvergel: I did the pass, hope I got all the comments

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
nvergel: However every other class follows that model

code/src/main/java/com/google/sps/storage/messaging/ContactStatus.java
nvergel: Why change getTimeWhenContactedSeconds to getTimeWhenContacted, isnt that making it less descriptive?

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserLocation.java
nvergel: But to address your comment, the reason why I used Long instead of long is because Objectify will set it to a non null value when stores. So a UUID would have the same purpose if you prefer that.

code/src/main/java/com/google/sps/storage/data-storage/Location.java
nvergel: Only two classes have get long/lat so I changed it for both to E7

code/src/main/java/com/google/sps/storage/data-storage/Location.java
nvergel: done

code/src/main/java/com/google/sps/storage/data-storage/Constants.java
nvergel: There are four constants now, I think they are properly formatted

code/src/main/java/com/google/sps/storage/data-storage/Location.java
bengaskagoogle: I almost wonder if the comment does the work here, because the code doesn't otherwise use lat/long to distinguish longitudeE7 against. I think the variable name and the name of the getter functions might be leaking too much information to the caller. 

Also, the multiplication isn't done in this class either, which makes it even more confusing. It's name is leaking assumptions about how the caller is constructing the object.

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
```diff
@@ -0,0 +1,54 @@
...
+import com.googlecode.objectify.annotation.Index;
+import java.time.Instant;
+import java.util.ArrayList;
+
+/**
+* This class stores the information needed to identify a negative user
+* at a specific place.
+*/
+@Entity
+public class NegativeUserPlace {
```
bengaskagoogle: I think the names Place/Location have a lot of overlap that confuses how they are different.

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserLocation.java
```diff
@@ -0,0 +1,55 @@
...
+/**
+* This class stores the information needed to identify a negative user
+* at a specific location (geographic coordinate)
+*/
+@Entity
+public class NegativeUserLocation {
+
+  // Location History provides geographic coordinates multiplied by 10^7 for more accuracy
+  private String  userId;
+  @Index private long longitudeE7;
```
bengaskagoogle: There is a lot of redundancy between NegativeUserLocation and Location. This might be a reasonable place for the NegativeUserLocation object to store a Location object rather than duplicating the fields in each class.

Is there a reason you're doing it this way? Same for Users vs NegativeUsers and Places.

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
nvergel: addressed during sync

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserLocation.java
nvergel: addressed during sync

code/src/main/java/com/google/sps/storage/data-storage/Business.java
```diff
@@ -0,0 +1,41 @@
...
+import java.time.Instant;
+import java.util.Optional;
+
+/**
+* This class is used to check if a business was contacted in the last week.
+*/
+@Entity
+public class Business {
+
+  @Id private String placeId;
```
bengaskagoogle: Also, I'm not sure if this has some issue with index, but a lot of these look to just be stores of value. If the values can't change it is worth making then final.

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
nvergel: A place is associated with a placeId and a location is associated with geographic coordinates. I do agree that it is a little confusing so I created a super class that's called TimeInterval. Now the implementation for Place.java and Location.java should be more obvious.

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserLocation.java
nvergel: We want to query against placeId, lat/lng, and time intervals for negative user but not positive users. This seperation allows us to make that distinction.

I would like to have NegativeUserLocation extend Location but it leads to the same problem. One class has @Index and the other does not.

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
nvergel: If you think this is a good solution I can do the same for NegativeUser and make a NegativeUserTimeInterval class

code/src/main/java/com/google/sps/storage/data-storage/Business.java
nvergel: I agree, this might be my lack of Java experience, but I was not sure if making them final would automatically set them to null and make it so we cannot change the value?

nvergel: I just looked up and it says final variables can be assigned once if they are left blank

nvergel: However, since Objectify requires a constructor with no parameters we would have to set placeId inside that empty constructor?

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserLocation.java
bengaskagoogle: Yeah, I was talking with Grady about this. I don't know what a clean solution is given the constraints of Objectify.

bengaskagoogle: So, unless Grady has a thought, this is probably resolved.

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
bengaskagoogle: It's usually best to favor composition (so storing a common object) over extending a class in Java. That might be the more correct path here.

code/src/main/java/com/google/sps/storage/data-storage/Business.java
bengaskagoogle: Hmm. I think then, unless we can figure out what Objectify is doing there and work around it, it's probably best to only make the variables you can final, ones that need to be passed in might necessarily be mutable.

code/src/main/java/com/google/sps/storage/data-storage/NegativeUserPlace.java
nvergel: done

code/src/main/java/com/google/sps/storage/data-storage/Location.java
nvergel: I think this is resolved?

PR #6, opened 2020-07-14
Cron job deletion

nvergel:


code/src/main/java/com/onlinecontacttracing/servlets/delete-data/DeletePositiveUserLocationsServlet.java
```diff
@@ -0,0 +1,19 @@
...
+@WebServlet("/delete-old-positive-user-locations")
+public class DeletePositiveUserLocationsServlet extends HttpServlet {
+
+  static final Logger log = Logger.getLogger(DeletePositiveUserLocationsServlet.class.getName());
+
+  @Override
+  public void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    OldDataDeleter.deleteOldData(OldDataDeleter.POSITIVE_USER_LOCATIONS, log);
+  }
+}
```
cynthia-m, line 19: these all have the same code structure but I guess there's not way to abstract that out?

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
```diff
@@ -0,0 +1,73 @@
...
+
+    // Fetch keys of old data
+    Iterable<Key<?>> oldDataKeys = ofy().load().type(oldDataDeleter.oldDataType)
+      .filter(oldDataDeleter.query, timeStampToGetOldData).keys();
+
+    // Save data type class name to report in log
+    String oldDataTypeName = oldDataDeleter.oldDataType.getSimpleName() + " class";
+
+    try {
+      // Check how many entries will be deleted
```
cynthia-m: is this comment necessary?

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
```diff
@@ -0,0 +1,73 @@
...
+      int numberOfDataToDelete = Iterables.size(oldDataKeys);
+
+      // Delete old data
```
cynthia-m: and this comment?

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
```diff
@@ -0,0 +1,73 @@
...
+      ofy().delete().keys(oldDataKeys);
+
+      // Log number of entries deleted
+      log.info( "Deleted " + numberOfDataToDelete + " entries from the " + oldDataTypeName);
+
+    } catch(Exception e) {
+      // Log error
```
cynthia-m: same for this comment and the one above?

cynthia-m (approved):

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/DeletePositiveUserLocationsServlet.java
nvergel: I think it is abstracted out with the OldDataDeleter class. Do you have any suggestion how it could be abstracted even more?

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
nvergel: updated comment

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/DeletePositiveUserLocationsServlet.java
cynthia-m: I guess not really, just thought I'd ask

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
```diff
@@ -0,0 +1,71 @@
...
+  * Otherwise the exception is caught and logged.
+  */
+  public static void deleteOldData(OldDataDeleter oldDataDeleter, Logger log) {
+    
+    /**
+    * To calculate the time for querying old data, this method
+    * subtracts maximum time a class is allowed to persist in memory
+    * from the current time by the 
+    */
+    long timeStampToGetOldData = Instant.now().getEpochSecond()-oldDataDeleter.maxTimeAllowedInStorage;
```
bengaskagoogle: nit: add spacing around operator.

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
```diff
@@ -0,0 +1,71 @@
...
+import com.onlinecontacttracing.storage.Constants;
+import com.google.common.collect.Iterables;
+import com.googlecode.objectify.Key;
+import java.util.logging.Logger;
+import java.time.Instant;
+
+/**
+* This class stores the information needed to query old data from a given class and delete it.
+*/
+public enum OldDataDeleter {
```
bengaskagoogle: This possibly sounds pedantic, but is there a regular DataDeleter or some specific logic in here that ties it to only deleting OldData? It seems like this class allows for an arbitrary query and time(?) to identify deletion items.

If so, this seems like a generic DataDeleter rather than anything specifically Old.

code/cron.yaml
```diff
@@ -0,0 +1,19 @@
+cron:
```
gradywardgoogle, line 1: I take it we were able to verify this works on the testing AppEngine instance? What manual testing did you perform there?

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
```diff
@@ -0,0 +1,71 @@
...
+    try {
+      // Check how many entries will be deleted to report in log
+      int numberOfDataToDelete = Iterables.size(oldDataKeys);
+
+      ofy().delete().keys(oldDataKeys);
+
+      // Log number of entries deleted
+      log.info( "Deleted " + numberOfDataToDelete + " entries from the " + oldDataTypeName);
+
+    } catch(Exception e) {
```
gradywardgoogle, line 40: I think you should check out the difference between this call with and without .now(), I think you'll need to add that in order to force this to be a synchronous call.

gradywardgoogle: https://stackoverflow.com/questions/28616680/why-now-objectify

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
gradywardgoogle: A slightly better structure would be to have an OldDataDeleter that is an abstract class, and has 3 abstract methods (getClass, getQueryString, and getMaxAge).

Then you could have 6 concrete implementations of this Abstract Class, which each implement those three methods (what you're doing here).

Then in the servlet you would say `new NotificationBatchDeleter().delete()`. and you'd have `class NotificationBatchDeleter extends OldDataDeleter`.

Does that make sense?

gradywardgoogle (commented):
Great start - one big structure comment.

code/cron.yaml
nvergel: Yes I tested it using our old project by having a simple server that would log every minute. Once that work I deployed it to our current project to see if it worked. Additionally, on the README there is a link to see the logs. Here it is: https://console.cloud.google.com/logs?_ga=2.119762766.714168666.1594744665-601550938.1590535110

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
nvergel: Okay, I will make it synchronous

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
nvergel: Makes sense but that sounds like a lot more code for the same functionality. Additionally it will be a lot of repeated code because the only difference between deletions is specifying the class, maximum age and query. Cynthia even commented that the current code could be abstracted even more since there is already repeated code (in the different servlets).

I will change the implementation if you want me to but I would rather work on other code (i.e. code not yet implemented).

nvergel: I agree

code/cron.yaml
gradywardgoogle: Awesome, thanks!

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/OldDataDeleter.java
gradywardgoogle: Happy to discuss this in detail, but inheritance is a better pattern here. The pattern you're using here is called static composition - you're encoding different types in static constants that you are calling enums. Why is inheritance better than static compositon (SC)?

* __Scalability__: If you need to add more per-type specificity, in SC need to add another argument (or even, a lambda function). In inheritance you'd just add another method, and you could support functions more gracefully.
* __Flexibility__: If you wanted at all custom behavior for any one of these behaviors, today you'd have to modify all of them (with a gross conditional). In an ineritance structure, you could just call super.fnName(), and then customize it.
* __Consistency__: Java is an object oriented programming language. We try to use its language features in a way that is consistent with the design principles of the language. This isn't an intended use case for enums.

I think Cynthia's point about code reuse on the servlet is a good one. But the way to solve that would be to add an abstract class with multiple concrete implementations, rather than having a statically composed set of servlets with a common set of functions.

gradywardgoogle: Also maybe I'm just reading into that last comment, but it sounds like you're a little bit frusturated here. While I can understand that, I want to push back a little and remind you about the distinction between engineering and code we talked about a while back.

In the future feel free to get early feedback on a code organization/strategy if you're unsure of it. This wasn't the right one in this case (which makes it a good learning opprotunity!) but if you're frusturated about iteration speed, the answer is earlier feedback, not worse code organization.

nvergel: Okay its done

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/NegativeUserPlacesDeleter.java
```diff
@@ -0,0 +1,60 @@
...
+  public long getMaxAge() {
+    return Instant.now().getEpochSecond() - Constants.NEGATIVE_USER_DATA_MAX_TIME;
+  }
+
+  /**
+  * This method deletes old data.
+  * If the query was successful, the number of entries deleted will be logged.
+  * Otherwise the exception is caught and logged.
+  */
+  public void delete(Logger log) {
```
gradywardgoogle: This method can be in the OldDataDeleter class, that's the purpose of the inheritance :)

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/PositiveUserPlacesDeleter.java
```diff
@@ -0,0 +1,60 @@
...
+  public String getQueryString() {
+    return "timeCreatedSeconds <";
+  }
+
+  /**
+  * To calculate the time for querying old data, this method
+  * subtracts maximum time a class is allowed to persist in memory
+  * from the current time by the 
+  */
+  public long getMaxAge() {
```
gradywardgoogle: If you modify this method to be "getRetentionDuration", you could just return the constant value, without doing Instant.now() here. You could do it in the abstract class.

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/DeleteNotificationBatchServlet.java
```diff
@@ -0,0 +1,19 @@
...
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+import java.util.logging.Logger;
+
+@WebServlet("/delete-old-notification-batch")
+public class DeleteNotificationBatchServlet extends HttpServlet {
+
+  static final Logger log = Logger.getLogger(DeleteNotificationBatchServlet.class.getName());
```
gradywardgoogle, line 13: If you moved the logger into each concrete instance of DataDeleter, that'd allow you to not have to pass it in here.

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/NegativeUserPlacesDeleter.java
nvergel: Makes sense, I think I wasn't zooming out and that's why I was hesitant to make the changes but this is clearly better. Sorry for the reluctance, it should be good now.

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/PositiveUserPlacesDeleter.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/delete-data/DeleteNotificationBatchServlet.java
nvergel: True, but every servlet needs to instantiate a logger based on its own class.

bengaskagoogle (approved):

PR #8, opened 2020-07-21
Send msg branch

cynthia-m:


code/pom.xml
```diff
@@ -8,6 +8,8 @@
   <version>1</version>
   <packaging>war</packaging>
 
+
```
bengaskagoogle: Lots of Whitespace. Make sure to do a clean-up pass before the real PR.

code/pom.xml
```diff
@@ -43,7 +79,52 @@
...
+        <artifactId>google-oauth-client-java6</artifactId>
+        <version>1.11.0-beta</version>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/com.google.oauth-client/google-oauth-client-jetty -->
+    <dependency>
+        <groupId>com.google.oauth-client</groupId>
+        <artifactId>google-oauth-client-jetty</artifactId>
+        <version>1.22.0</version>
+    </dependency>
+    <!-- https://mvnrepository.com/artifact/com.google.api.client/google-api-client-util -->
```
bengaskagoogle: Is this commented out?

code/src/main/java/com/onlinecontacttracing/messaging/LocalityResource.java
```diff
@@ -0,0 +1,73 @@
...
+  }
+  public String getEnglishTranslation() {
+    return englishTranslation.concat(helpLink);
+  }
+
+  public String getSpanishTranslation() {
+    return spanishTranslation.concat(helpLink);
+  }
+
+
```
bengaskagoogle: Whitespace

code/src/main/java/com/onlinecontacttracing/messaging/LocalityResource.java
```diff
@@ -0,0 +1,73 @@
...
+  VERMONT("https://www.healthvermont.gov/response/coronavirus-covid-19", "Here's your state's help link: ", "Aquí está el enlace de ayuda de su estado: "),
+  VIRGINIA("https://www.vdh.virginia.gov/coronavirus/", "Here's your state's help link: ", "Aquí está el enlace de ayuda de su estado: "),
+  WASHINGTON("https://www.doh.wa.gov/Emergencies/Coronavirus", "Here's your state's help link: ", "Aquí está el enlace de ayuda de su estado: "),
+  WEST_VIRGINIA("https://dhhr.wv.gov/COVID-19/Pages/default.aspx", "Here's your state's help link: ", "Aquí está el enlace de ayuda de su estado: "),
+  WISCONSIN("https://www.dhs.wisconsin.gov/covid-19/index.htm", "Here's your state's help link: ", "Aquí está el enlace de ayuda de su estado: "),
+  WYOMING("https://health.wyo.gov/publichealth/infectious-disease-epidemiology-unit/disease/novel-coronavirus/", "Here's your state's help link: ", "Aquí está el enlace de ayuda de su estado: ");
+  
+  private String helpLink;
+  private String englishTranslation;
+  private String spanishTranslation;
```
bengaskagoogle: I think you can make this final

code/src/main/java/com/onlinecontacttracing/messaging/LocalityResource.java
```diff
@@ -0,0 +1,73 @@
+package com.onlinecontacttracing.messaging;
+
+public enum LocalityResource implements HasEnglishTranslation, HasSpanishTranslation{
+  US("https://www.cdc.gov/coronavirus/2019-ncov/index.html", "Here's the national help link: ", "Aquí está el enlace de ayuda nacional: "),
```
bengaskagoogle, line 4: So, The big thing to note here is that everything is repeated except the state link value.That indicates to me that this can be factored out into a common variable rather than stored in each enum.

code/src/main/java/com/onlinecontacttracing/storage/data-storage/PositiveUser.java
```diff
@@ -62,14 +70,14 @@ public boolean userCanStillSendEmails() {
...
 
-  public boolean userCanMakeMoreDraftsAfterBeingFlagged() {
-    if (attemptedEmailDrafts > 0) {
-      attemptedEmailDrafts--;
-      return true;
-    } else {
-      return false;
-    }
-  }
+  // public boolean userCanMakeMoreDraftsAfterBeingFlagged() {
```
bengaskagoogle: Make sure that this is either completed or removed at the point of the PR. I think it should be removed form this specific one though as I think it's not necessary for what we're discussing here. Remember, keep PRs small.

code/pom.xml
```diff
@@ -8,6 +8,8 @@
   <version>1</version>
```
bengaskagoogle: High Level comment. 

1) This PR is humongous, many of these files could be their own PR. Don't bother breaking it up now, I think, but keep this in mind for future PR. 
2) Please test.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,97 @@
+package com.onlinecontacttracing.messaging;
+
+import com.onlinecontacttracing.messaging.LocalityResource;
+import com.onlinecontacttracing.storage.CustomizableMessage;
+import com.onlinecontacttracing.storage.PositiveUser;
+import com.onlinecontacttracing.storage.PotentialContact;
+import com.onlinecontacttracing.messaging.filters.CheckMessagesForFlags;
+import java.util.List;
+
+
```
bengaskagoogle: Whitespace. You should have some space between static imports and regular imports, but otherwise put them together.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,97 @@
...
+  //if flags are triggered, returns message and list of errors, else returns message
+  public List<String> statusListToShowUser(String messageLanguage) {
+    String translatedResourceMessage;
+    String translatedSystemMessage;
+    user.incrementAttemptedEmailDrafts();
+    checkForFlags();
+    
+    if (messageLanguage.equals("SP")) {
+    translatedResourceMessage = localityResource.getEnglishTranslation();
```
bengaskagoogle: Indentation is wrong

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,97 @@
...
+    translatedSystemMessage = systemMessage.getEnglishTranslation();
+    }
+    else {
+      translatedResourceMessage = localityResource.getEnglishTranslation();
+      translatedSystemMessage = systemMessage.getEnglishTranslation();
+    }
+    
+    //if any flags are triggered, do not want to include the user's message in the message sent to the backend
+    if (statusMessage.size() != 0){
```
bengaskagoogle: statusMessage.size() > 0 

makes more explicit what you intend for this here.

bengaskagoogle: Also, what is the userMessage in the case that statusMessage is size 0? If you can make that be equivalent to an empty string, then you can just concat without this check and clean things up slightly.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,97 @@
...
+      compiledMessage = translatedSystemMessage.concat(translatedResourceMessage);
+    } else {
+      compiledMessage = (translatedSystemMessage.concat(userMessage).concat(translatedResourceMessage);
+    }
+    statusMessage.add(translatedSystemMessage.concat(userMessage).concat(translatedResourceMessage));
+    return statusMessage;
+
```
bengaskagoogle: Whitespace

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,97 @@
...
+      translatedSystemMessage = systemMessage.getEnglishTranslation();
+    }
+    
+    //if any flags are triggered, do not want to include the user's message in the message sent to the backend
+    if (statusMessage.size() != 0){
+      compiledMessage = translatedSystemMessage.concat(translatedResourceMessage);
+    } else {
+      compiledMessage = (translatedSystemMessage.concat(userMessage).concat(translatedResourceMessage);
+    }
+    statusMessage.add(translatedSystemMessage.concat(userMessage).concat(translatedResourceMessage));
```
bengaskagoogle: I note you don't do the check here as well. 

What is the compiledMessage being used as? It's a little concerning that we have these two fields that are almost exactly the same.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,97 @@
...
+  private String compiledMessage;
+  //for frontend display
+  private List<String> statusMessage = new ArrayList<String> ();
+
+
+  public MessageServlet(SystemMessage systemMessage, LocalityResource localityResource, CustomizableMessage customizableMessage, PositiveUser user) {
+    this.systemMessage = systemMessage;
+    this.localityResource = localityResource;
+    this.customizableMessage =  customizableMessage;
+    this.userMessage = customizableMessage.getMessage();
```
bengaskagoogle: In general, you don't want to do work in the constructor. Even calling get like this. Probably just store the CustomizableMessage and do the getMessage() call below.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,97 @@
...
+      compiledMessage = (translatedSystemMessage.concat(userMessage).concat(translatedResourceMessage);
+    }
+    statusMessage.add(translatedSystemMessage.concat(userMessage).concat(translatedResourceMessage));
+    return statusMessage;
+  
+  }
+
+  public String compileMessage () {
+    //returns message with no userMessage if there are errors, else return message with userMessage
+    if (statusMessages.size() > 1) {
```
bengaskagoogle: Is this an attempt at caching?

If so, don't do this. Just general construct the message each time. Or else, fully cache the message and get rid of this weird choice logic.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,97 @@
...
+      compiledMessage = translatedSystemMessage.concat(translatedResourceMessage);
+    } else {
+      compiledMessage = (translatedSystemMessage.concat(userMessage).concat(translatedResourceMessage);
+    }
+    statusMessage.add(translatedSystemMessage.concat(userMessage).concat(translatedResourceMessage));
+    return statusMessage;
+  
+  }
+
+  public String compileMessage () {
```
bengaskagoogle: This difference between statusMessage and compiledMessage is very, very confusing. Here is a good place to rethink how you've structured things.

code/src/main/java/com/onlinecontacttracing/messaging/MessagingSetup.java
```diff
@@ -0,0 +1,115 @@
...
+import javax.mail.Multipart;
+import javax.mail.Session;
+import javax.mail.internet.InternetAddress;
+import javax.mail.internet.MimeBodyPart;
+import javax.mail.internet.MimeMessage;
+import javax.mail.internet.MimeMultipart;
+import java.io.ByteArrayOutputStream;
+
+/** Source Code: https://developers.google.com/gmail/api/guides/sending*/
+public class MessagingSetup {
```
bengaskagoogle, line 43: Is this taken unmodified from the startup guide? Also a very useful one to check in independently.

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
```diff
@@ -0,0 +1,94 @@
...
+
+import com.google.api.client.auth.oauth2.Credential;
+import com.google.api.client.googleapis.auth.oauth2.GoogleAuthorizationCodeFlow;
+import com.google.api.client.extensions.java6.auth.oauth2.AuthorizationCodeInstalledApp;
+import com.google.api.client.googleapis.auth.oauth2.GoogleClientSecrets;
+import com.google.api.client.googleapis.javanet.GoogleNetHttpTransport;
+import com.google.api.client.http.javanet.NetHttpTransport;
+import com.google.api.client.json.JsonFactory;
+import com.google.api.client.json.jackson2.JacksonFactory;
+// import com.google.api.client.util.FileDataStoreFactory;
```
bengaskagoogle: nit: remove.

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
```diff
@@ -0,0 +1,94 @@
...
+  private String emailBody;
+  private MessageServlet messageObject;
+  private ArrayList<PotentialContact> contactsList;
+  private Gmail service;
+  private PositiveUser user;
+
+  private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+  private static final String TOKENS_DIRECTORY_PATH = "tokens";
+  private static final List<String> SCOPES = Collections.singletonList(GmailScopes.GMAIL_SEND);
+  private static final String CREDENTIALS_FILE_PATH = "/credentials.json";//rename to be likek authconfigurationdata, also need to remove from github
```
bengaskagoogle: nit: typo, also move above the code, so

// rename to be like ...
private static final String ...

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
```diff
@@ -0,0 +1,94 @@
...
+  // final NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
```
bengaskagoogle: I'm guessing this is something still being worked on.

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
```diff
@@ -0,0 +1,94 @@
...
+  
+  public SendEmails (SystemMessage systemMessage, LocalityResource localityResource, CustomizableMessage customizableMessage, String emailSubject, String messageLanguage, ArrayList<PotentialContact> contactsList, PositiveUser user) {
+    messageObject = new MessageServlet(systemMessage, localityResource, customizableMessage, user);
+    emailBody = messageObject.compileMessage(messageObject.statusListToShowUser(messageLanguage));
```
bengaskagoogle: use this.* for each of them. Just do it to be explicit. Same for the other constructors.

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
```diff
@@ -0,0 +1,94 @@
...
+  private Gmail service;
+  private PositiveUser user;
+
+  private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+  private static final String TOKENS_DIRECTORY_PATH = "tokens";
+  private static final List<String> SCOPES = Collections.singletonList(GmailScopes.GMAIL_SEND);
+  private static final String CREDENTIALS_FILE_PATH = "/credentials.json";//rename to be likek authconfigurationdata, also need to remove from github
+  // final NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
+  
+  public SendEmails (SystemMessage systemMessage, LocalityResource localityResource, CustomizableMessage customizableMessage, String emailSubject, String messageLanguage, ArrayList<PotentialContact> contactsList, PositiveUser user) {
```
bengaskagoogle: Since this is a class its more likely to be named EmailSender or something. Nouns for classes, verbs for methods.

code/src/main/java/com/onlinecontacttracing/messaging/LocalityResource.java
bengaskagoogle: Also, what led to this big enum structure like this? It feels slightly unwieldy to me. Though it's difficult with a big list. 

How did you come upon this structure in particular?

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
```diff
@@ -0,0 +1,94 @@
...
+    MimeMessage email;
+    for(PotentialContact contactName : contactsList) {
+      // email = MessagingSetup.createEmail(contactName.getEmail(), user.getUserEmail(), emailSubject, emailBody);
+      // sendMessage(service, user.getUserId(), email);
+    }
+  }
+
+  private static Credential getCredentials(final NetHttpTransport HTTP_TRANSPORT, PositiveUser user) throws IOException {
+        // Load client secrets
+        InputStream in = SendEmails.class.getResourceAsStream(CREDENTIALS_FILE_PATH);//create a class authAPI config constants, have this as a string, eliminate input reader
```
bengaskagoogle: Move above.

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
```diff
@@ -0,0 +1,94 @@
...
+        if (in == null) {
+            throw new FileNotFoundException("Resource not found: " + CREDENTIALS_FILE_PATH);
+        }
+        GoogleClientSecrets clientSecrets = GoogleClientSecrets.load(JSON_FACTORY, new InputStreamReader(in));
+
+      
+        TokenResponse tokenResponse = new TokenResponse().setScope(SCOPES.get(0));
+        return new GoogleAuthorizationCodeFlow.Builder(
+               HTTP_TRANSPORT, JSON_FACTORY, clientSecrets, SCOPES).build().createAndStoreCredential(tokenResponse, user.getUserId());
+        //  = new GoogleAuthorizationCodeFlow.Builder(
```
bengaskagoogle: Presumably you're still working on this?

code/src/main/java/com/onlinecontacttracing/messaging/SystemMessage.java
```diff
@@ -0,0 +1,23 @@
+package com.onlinecontacttracing.messaging;
+
+public enum SystemMessage implements HasEnglishTranslation, HasSpanishTranslation  {
+  VERSION_1("Hello, you are receiving this message because someone you know has tested positive for COVID-19. Please consider quarantining and self-isolating for 14 weeks in addition to getting tested.",
```
bengaskagoogle, line 4: This one too. How are you expecting this code to be consumed that you're using an enum here?

code/src/main/java/com/onlinecontacttracing/messaging/credentials.json
```diff
@@ -0,0 +1 @@
+{"web":{"client_id":"1080865471187-u1vse3ccv9te949244t9rngma01r226m.apps.googleusercontent.com","project_id":"covid-catchers-fixed-gcp","auth_uri":"https://accounts.google.com/o/oauth2/auth","token_uri":"https://oauth2.googleapis.com/token","auth_provider_x509_cert_url":"https://www.googleapis.com/oauth2/v1/certs","client_secret":"LpbG3OQ9SqSckXfkWikxXFUn","redirect_uris":["https://covid-catchers-fixed-gcp.ue.r.appspot.com"],"javascript_origins":["https://8080-ac896ae1-5f0c-45b5-8dfe-19fa4d3d8699.us-east1.cloudshell.dev","https://covid-catchers-fixed-gcp.ue.r.appspot.com"]}}
```
bengaskagoogle: I don't know for sure, but this definitely looks like something that shouldn't be checked in directly.

code/src/main/java/com/onlinecontacttracing/messaging/filters/CheckMessagesForFlags.java
```diff
@@ -0,0 +1,37 @@
...
+import com.onlinecontacttracing.storage.CustomizableMessage;
+import com.onlinecontacttracing.storage.PositiveUser;
+import com.onlinecontacttracing.storage.PotentialContact;
+import java.util.ArrayList;
+import java.util.List;
+import com.onlinecontacttracing.messaging.filters.FlaggingFilter;
+
+//checks all the filters
+public class CheckMessagesForFlags {
+  private final static List<FlaggingFilter> listOfFilters = new ArrayList<FlaggingFilter> () {{
```
bengaskagoogle: I think this is good and makes a lot of sense.

code/src/main/java/com/onlinecontacttracing/messaging/filters/CustomizeMessageTriesFlaggingFilter.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.ArrayList;
+import com.onlinecontacttracing.messaging.filters.FlaggingFilter;
+import com.onlinecontacttracing.storage.PositiveUser;
+import com.onlinecontacttracing.storage.PotentialContact;
+import java.lang.Exception;
+
+//Checks if user has attempted to submit a message too many times
+public class CustomizeMessageTriesFlaggingFilter implements FlaggingFilter{
+  private int maxNumberOfTries = 3;
```
bengaskagoogle: for a hard coded value like this, it should be a private static final in and should be named MAX_NUMBER_OF_TRIES

code/src/main/java/com/onlinecontacttracing/messaging/filters/ProfanityFlaggingFilter.java
```diff
@@ -0,0 +1,64 @@
...
+
+//Checks if message contains any profanity
+public class ProfanityFlaggingFilter implements FlaggingFilter{
+  private static ArrayList<String> listOfProfanityIndicators = = new ArrayList<String> () {{
+    add("fuck");
+    add("shit");
+    add("bitch");
+    add("crap");
+    add("damn");
+    add("goddamn");
```
bengaskagoogle: This is mostly fine, but most likely you'd actually define some dictionary file you read in. This makes it much easier to expand upon.

code/src/main/java/com/onlinecontacttracing/messaging/filters/NumberOfMessagesFlaggingFilter.java
```diff
@@ -0,0 +1,20 @@
...
+import com.onlinecontacttracing.messaging.filters.FlaggingFilter;
+import com.onlinecontacttracing.storage.PositiveUser;
+import com.onlinecontacttracing.storage.PotentialContact;
+import java.lang.Exception;
+
+public class NumberOfMessagesFlaggingFilter implements FlaggingFilter{
+  private static int limitNumOfMessages = 100;
+
+  public boolean passesFilter(PositiveUser positiveUser, String message) {
+    if (positiveUser.getNumberOfEmailsSent() > limitNumOfMessages) {
```
bengaskagoogle: If you have a guard that just returns a boolean, you can just return the expression itself.

So, 

return !positiveUser.getNumberOfEmailsSent() > limitNumOfMesssages;

code/src/main/java/com/onlinecontacttracing/messaging/filters/LinkFlaggingFilter.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+    return true;
+  }
+
+  public String errorMessageToUser() {
+    return "We think there might be a URL or link in your message. Please remove them and try again!";
+  }
+
+  
+
```
bengaskagoogle: Whitespace.

code/src/main/java/com/onlinecontacttracing/messaging/filters/LengthFlaggingFilter.java
```diff
@@ -0,0 +1,21 @@
...
+import com.onlinecontacttracing.storage.PositiveUser;
+import com.onlinecontacttracing.storage.PotentialContact;
+import java.lang.Exception;
+
+//Checks that the length of the message is within limits
+public class LengthFlaggingFilter implements FlaggingFilter{
+  private static int limitNumOfCharacters = 500;
+  
+  public boolean passesFilter(PositiveUser positiveUser, String message) {
+    if (message.length() > limitNumOfCharacters) {
```
bengaskagoogle: Same as comment below. You can just return message.length() <= limitNumOfCharacters

code/pom.xml
cynthia-m: yup I removed it

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
cynthia-m: The idea was that statusMessage is to be displayed to the user in the frontend, and compiledMessage will be used in the backend. For the frontend message, we still want to display the user's message to them even if there are errors, but for the backend, if there are flags, we wouldn't want to include the message in the email. But I think these variable names are probably not a descriptive as they could be so I'll fix that

cynthia-m: and also maybe add more comments to help clarify

code/src/main/java/com/onlinecontacttracing/messaging/MessagingSetup.java
cynthia-m: Yup almost exactly, how should I source this?

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
cynthia-m: yup

code/src/main/java/com/onlinecontacttracing/messaging/SendEmails.java
cynthia-m: yup, this is part of the "need service account" section

code/src/main/java/com/onlinecontacttracing/messaging/credentials.json
cynthia-m: yes, I need to remove this which I'll do right now

code/src/main/java/com/onlinecontacttracing/messaging/filters/CustomizeMessageTriesFlaggingFilter.java
cynthia-m: gotcha

code/src/main/java/com/onlinecontacttracing/messaging/LocalityResource.java
cynthia-m: It was something Grady and I had discussed when we were writing the DD, I can segment out a bit of it I think but I don't think I can completely eliminate all the enum values unless you've got any ideas?

code/pom.xml
```diff
@@ -33,6 +33,16 @@
         <artifactId>appengine-api-1.0-sdk</artifactId>
         <version>1.9.59</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.apis</groupId>
+      <artifactId>google-api-services-gmail</artifactId>
+      <version>v1-rev20200601-1.30.10</version>
+    </dependency>
+    <dependency>
```
gradywardgoogle: Similar to below plz include the link to the maven repository here.

code/pom.xml
```diff
@@ -33,6 +33,16 @@
         <artifactId>appengine-api-1.0-sdk</artifactId>
         <version>1.9.59</version>
     </dependency>
+    <dependency>
+      <groupId>com.google.apis</groupId>
+      <artifactId>google-api-services-gmail</artifactId>
+      <version>v1-rev20200601-1.30.10</version>
```
gradywardgoogle: Can you try to sort these alphabetically, so it's easier to check for duplicates/inclusion?

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.security.GeneralSecurityException;
+import java.util.Collections;
+import java.util.List;
+import java.util.ArrayList;
+import javax.mail.internet.MimeMessage;
+
+public class EmailSender {
```
gradywardgoogle, line 38: Documentation plz

gradywardgoogle, line 38: So, Let's break this class up into two parts:
EmailSender takes in a credential and a Positive User, and constructs a GMail Service and stores the user.

Message takes in SystemMessage systemMessage, LocalityResource localityResource, CustomizableMessage customizableMessage and MessageLanguage.

Then EmailSender has one method sendEmails(Message), and the Message class has one method getAsString().

The reason this is a better pattern is that the message properties are better captured as a part of a message object, and the sender properties are better captured as a part of the sender object.

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,100 @@
...
+    this.emailSubject= emailSubject;
+    this.contactsList = contactsList;
+    // service = new Gmail.Builder(HTTP_TRANSPORT, JSON_FACTORY, getCredentials(HTTP_TRANSPORT, user))
+    //             .setApplicationName(APPLICATION_NAME)
+    //             .build();
+    this.user = user;
+  }
+
+  public void setMessageObject(SystemMessage systemMessage, LocalityResource localityResource, CustomizableMessage customizableMessage, PositiveUser user) {
+    this.messageObject = new MessageServlet(systemMessage, localityResource, customizableMessage, user);
```
gradywardgoogle: I don't think this should be a servlet, just a Message.

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,100 @@
...
+        return new GoogleAuthorizationCodeFlow.Builder(
+               HTTP_TRANSPORT, JSON_FACTORY, clientSecrets, SCOPES).build().createAndStoreCredential(tokenResponse, user.getUserId());
+        //  = new GoogleAuthorizationCodeFlow.Builder(
+        //         HTTP_TRANSPORT, JSON_FACTORY, clientSecrets, SCOPES)
+        //         .setDataStoreFactory(new FileDataStoreFactory(new java.io.File(TOKENS_DIRECTORY_PATH)))
+        //         .setAccessType("offline")
+        //         .build();
+        // LocalServerReceiver receiver = new LocalServerReceiver.Builder().setPort(8888).build();
+        // return new AuthorizationCodeInstalledApp(flow, receiver).authorize("user");
+    }
```
gradywardgoogle: If I understand correctly, this is still getting the credentials for the sending user, rather than for our service account. Is that right?

code/src/main/java/com/onlinecontacttracing/messaging/LocalityResource.java
gradywardgoogle: I agree with Ben that we don't want all of the duplicate data. For example, we know that "Here's your state's help link" doesn't change relative to the state. So that should just be a static variable used in the getEnglishTranslation method. The State Health links on the otherhand do varry by state, so they do need to be part of the enum construction.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,95 @@
...
+import com.google.appengine.api.datastore.Query.SortDirection;
+import com.google.appengine.api.datastore.Key;
+import com.google.appengine.api.datastore.Projection;
+import java.util.Collections;
+import java.util.List;
+import java.util.ArrayList;
+import java.io.IOException;
+
+@WebServlet("/messaging")
+public class MessageServlet extends HttpServlet {
```
gradywardgoogle: So same as above, it seem strange to me to see this as a servlet, I think it's just a java class.

gradywardgoogle: We never use the doGet, so that's a good signal this should just be a class called Message.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,95 @@
...
+    CheckMessagesForFlags flagChecker = new CheckMessagesForFlags();
+    this.messagesForFrontendDisplay = CheckMessagesForFlags.findTriggeredFlags(flagChecker, this.user, this.userMessage);
+  }
+
+  //if flags are triggered, returns message and list of errors, else returns message
+  public ArrayList<String> compileMessages(String messageLanguage) {
+    String translatedResourceMessage;
+    String translatedSystemMessage;
+    this.user.incrementAttemptedEmailDrafts();
+    checkForFlags();
```
gradywardgoogle: Let's move this logic to check for validity to the MessageSender class.

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,95 @@
...
+    
+    if (messageLanguage.equals("SP")) {
```
gradywardgoogle: I'd prefer we represent the message language as an enum:

enum Language {
  ENGLISH, 
  SPANISH;
}

code/src/main/java/com/onlinecontacttracing/messaging/MessageServlet.java
```diff
@@ -0,0 +1,95 @@
...
+      translatedResourceMessage = localityResource.getEnglishTranslation();
```
gradywardgoogle: nit: I think this should be spanish.

code/src/main/java/com/onlinecontacttracing/messaging/filters/FileReader.java
```diff
@@ -0,0 +1,23 @@
+package com.onlinecontacttracing.messaging.filters;
+
+import java.io.File;
+import java.io.FileNotFoundException;
+import java.util.Scanner;
+import java.util.ArrayList;
+
+public class FileReader {
+  public static ArrayList<String> getListFromFile(String fileName) {
+    ArrayList<String> listOfIndicators = new ArrayList<String> ();
```
gradywardgoogle, line 10: this variable is named unintuitively. Isn't this just the lines?

gradywardgoogle (commented):
Happy to do detailed review when ready!

PR #9, opened 2020-07-22
add servlet infrastructure

nvergel:


code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
```diff
@@ -0,0 +1,21 @@
...
+public class GetNegativeUserInfoServlet extends HttpServlet {
+
+  @Override
+  public void doPost(HttpServletRequest request, HttpServletResponse response) throws IOException {
+    try {
+      request.getRequestDispatcher("/get-negative-user-calendar-info").forward(request,response);
+    } catch(Exception e) {
+      e.printStackTrace();
+    }
+    System.out.println("Negative User done getting info");
```
Josiah-Caldwell: How about "Done getting negative user info"

Josiah-Caldwell: Also I don't think it should print that if there is an exception

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetPositiveUserInfoServlet.java
```diff
@@ -0,0 +1,31 @@
...
+    } catch(Exception e) {
+      e.printStackTrace();
+    }
+    // Wait for servlets to finish executing?
+    try {
+      request.getRequestDispatcher("/merge-positive-user-contacts").forward(request,response);
+    } catch(Exception e) {
+      e.printStackTrace();
+    }
+    System.out.println("Positive User done getting info");
```
Josiah-Caldwell: Same comments as for negative user, reformat and move into the try block

code/src/main/webapp/JavaScript/script.js
```diff
@@ -172,12 +172,17 @@ function attachSignin(element, negativeUser) {
...
-    });
+    if (negativeUser) {
+      const request = new Request('/get-negative-user-info', {method: 'POST', body: params});
+      fetch(request).then(() => {
+        PAGE_CONTROLLER.show('notification');
+      });
+    } else {
+      const request = new Request('/get-positve-user-info', {method: 'POST', body: params});
+      fetch(request).then(() => {
+        // Redirect to next page
```
Josiah-Caldwell: Is this redirect to the custom message page?

nvergel: No this is redirect to approve lists

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
nvergel: This print statements are just for debugging. Once we implement the actual servlets we can remove them

Josiah-Caldwell (approved):

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
gradywardgoogle: Let's either log, or put it in the response. System.out.println will be hard to find in the stack traces, where as a log will be obvious and a response makes it easy to validate.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetPositiveUserInfoServlet.java
```diff
@@ -0,0 +1,85 @@
...
+    updateUser(request.getParameter("idToken"));
+
+    // Execute runnable to get people data
+    ArrayList<PotentialContact> contactsFromPeople = new ArrayList<PotentialContact>();
+    Thread peopleInfo = new Thread(new GetPeopleData(contactsFromPeople));
+    peopleInfo.start();
+
+    // Forward to servlet to retrieve calendar data
+    try {
+      request.getRequestDispatcher("/get-positive-user-calendar-info").forward(request,response);
```
gradywardgoogle: Similar to what you did with getPeopleData, rather than doing a servlet forward here, let's create a runnable that we can call from both servlets.

gradywardgoogle (commented):
This will probably all have to wait on the Authentication code going in, but I like the general structure here.

I think the pattern of abstracting the fetch logic to a runnable and then calling one or more runnables from the servlet is a reasonable pattern, but we'll want to do it consistently in any context where we've got multiple calls - let's not mix forwarding and threading.

Also: we'll have to think about how authentication mixes into that story. We'll probably want to do the authentication for all of the requests before we run any multi-threaded code, and we won't want the multi-threaded code to do authentication.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
```diff
@@ -0,0 +1,62 @@
...
+    Credential credential;
+
+    public GetCalendarData(Credential credential) {
+      this.credential = credential;
+    }
+
+    public void run() {
+      // Get contacts from people api
+    }
+  }
```
gradywardgoogle: Let's move the actions into their own classes.

As a side note, I assume that there isn't output from these actions, right? They just store the output in datastore? If you do have to create an action that has a return type, I've got a few ideas.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
```diff
@@ -0,0 +1,62 @@
...
+import java.util.Optional;
+
+@WebServlet("/get-negative-user-info")
+public class GetNegativeUserInfoServlet extends CheckForApiAuthorizationServlet {
+
+  String getServletURIName() {
+    return "get-negative-user-info";
+  }
+  
+  void useCredential(Credential credential) {
```
gradywardgoogle: Lets change the interface here to `Optional<Exception> executePostForAuthorizedUser(String userId, Credential credential)`

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
```diff
@@ -0,0 +1,62 @@
...
+    Thread contactInfo = new Thread(new GetCalendarData(credential));
+
+    contactInfo.start();
+
+    try {
+      contactInfo.join();
+    } catch(Exception e) {
+      e.printStackTrace();
```
gradywardgoogle: We should return an error up to the abstract class, and then the abstract class should be able to create a 500 error response with an informative message from that.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    System.out.println("Negative User done getting info");
```
gradywardgoogle: Either drop this, or log it instead of S.O.PLN

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
```diff
@@ -0,0 +1,62 @@
...
+    public GetCalendarData(Credential credential) {
+      this.credential = credential;
+    }
+
+    public void run() {
+      // Get contacts from people api
+    }
+  }
+
+  public void updateUser(String userId) {
```
gradywardgoogle: unused?

gradywardgoogle (commented):
Generally looks good, some small structural changes.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
nvergel: This is might be something to address in the auth PR

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
nvergel: Dropped

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
nvergel: part of the abstract class, used there

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
nvergel: None of them return anything, however I pass an ArrayList to one of them that I later use outside of the runnable.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
nvergel: Okay, I will wait for auth PR to be fully merged before I work on this

nvergel:
I addressed everything I could, the rest needs the auth branch to be merged

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
nvergel: Done!

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -61,19 +61,27 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+
     } catch (FileNotFoundException e) {
       log.warning("credentials.json not found");
       response.sendRedirect("/?page=login&error=FileError");
     } catch (GeneralSecurityException e) {
       log.warning("http transport failed, security error");
-      response.sendRedirect("/?page=login&error=TransportError");
-    } catch(Exception e) { //don't expect any other error
+      response.sendRedirect("/?page=login&error=GeneralError");
+    } catch (InterruptedException e) {
```
bengaskagoogle: If you're not handling it at all different from the general exception it doesn't make sense to add a new catch.

Did you intend to log something here? Or redirect to a different page?

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -61,19 +61,27 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
+      response.sendRedirect("/?page=login&error=GeneralError");
+    } catch(Exception e) { // don't expect any other error
       log.warning("exception occurred");
```
bengaskagoogle: Probably log the actually exception as well. You can do so with

log.warning(e.toString());

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -113,18 +121,15 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
   /**
```
bengaskagoogle: Update the docs too. Probably remove them since it is private now.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetCalendarDataForNegativeUser.java
```diff
@@ -0,0 +1,15 @@
+package com.onlinecontacttracing.authentication;
+
+import com.google.api.client.auth.oauth2.Credential;
+
+class GetCalendarDataForNegativeUser implements Runnable {
```
bengaskagoogle: I'm actually guessing there will be no difference in how you get the calendar data for a positive or negative user. Can you tell me what you think will differ to the point that it requires two different classes?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
```diff
@@ -0,0 +1,48 @@
...
+public class GetNegativeUserInfoServlet extends CheckForApiAuthorizationServlet {
+
+  String getServletURIName() {
+    return "/get-negative-user-info";
+  }
+  
+  void useCredential(Credential credential, HttpServletResponse response) throws IOException, InterruptedException {
+    Thread contactInfo = new Thread(new GetCalendarDataForNegativeUser(credential));
+
+    contactInfo.start();
```
bengaskagoogle: Why exactly are you kicking off a thread than immediately joining it?

I'm guessing this is an attempt to spin off a thread and then wait for it to finish?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetCalendarDataForPositiveUser.java
```diff
@@ -0,0 +1,15 @@
+package com.onlinecontacttracing.authentication;
+
+import com.google.api.client.auth.oauth2.Credential;
+
+class GetCalendarDataForPositiveUser implements Runnable {
```
bengaskagoogle: Will there be anything special about a PostiveUser that can't just be CalendarDataForUser instead?

bengaskagoogle: Also, Classes are nouns, not verbs. Methods are verbs.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
bengaskagoogle: And then the data isn't used?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetPositiveUserInfoServlet.java
```diff
@@ -0,0 +1,61 @@
...
+import static com.googlecode.objectify.ObjectifyService.ofy;
+import com.onlinecontacttracing.storage.PositiveUser;
+import java.util.Optional;
+import java.util.ArrayList;
+
+@WebServlet("/get-positve-user-info")
+public class GetPositiveUserInfoServlet extends CheckForApiAuthorizationServlet {
+
+  String getServletURIName() {
+    return "/get-positve-user-info";
```
bengaskagoogle: Typo here and above.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetPositiveUserInfoServlet.java
```diff
@@ -0,0 +1,61 @@
...
+
+    peopleInfo.start();
+    contactInfo.start();
+
+    
+    peopleInfo.join();
+    contactInfo.join();
+
+    // Load PositiveUserContacts from objectify
+    // call mergeContactListsFromPeopleAPI(contactsFromPeople)
```
bengaskagoogle: Incomplete code again.

code/src/main/java/com/onlinecontacttracing/storage/data-storage/NegativeUser.java
```diff
@@ -13,13 +13,20 @@
...
+  private long firstLoginSeconds;
+  private long lastLoginSeconds;
 
   // Objecify requires one constructor with no parameters
   private NegativeUser() {}
   
   public NegativeUser(String id, String email) {
     userId = id;
     userEmail = email;
+    firstLoginSeconds = lastLoginSeconds = Instant.now().getEpochSecond();
```
bengaskagoogle: Get rid of this chain assignment.

code/src/main/java/com/onlinecontacttracing/storage/data-storage/NegativeUser.java
```diff
@@ -13,13 +13,20 @@
...
+  }
+
+  public void setLastLogin() {
+    firstLoginSeconds = Instant.now().getEpochSecond();
```
bengaskagoogle: This probably is updating the wrong field. You probably want to update lastLoginInSeconds

code/src/main/java/com/onlinecontacttracing/storage/data-storage/NegativeUser.java
```diff
@@ -30,8 +37,16 @@ public String getUserEmail() {
...
+  }
+
+  public long getLastLoginInSeconds() {
+    return lastLoginSeconds;
+  }
+  
   @Override
   public String toString() {
-    return "Negative User- ID:" + userId + ", email:" + userEmail;
+    return "Negative User- ID: " + userId + ", email: " + userEmail;
```
bengaskagoogle: nit: Probably should be "User-ID" not "User- ID"

Also, do you know about String.format?

bengaskagoogle: String.format("Negative User-ID: %s, email: %s,", userId, userEmail); 

Might be easier.

code/src/main/java/com/onlinecontacttracing/storage/data-storage/PositiveUser.java
```diff
@@ -73,7 +73,7 @@ public boolean userCanMakeMoreDraftsAfterBeingFlagged() {
 
   @Override
   public String toString() {
-    String person = "Negative User- ID:" + userId + ", email:" + userEmail + "\n";
+    String person = "Negative User- ID: " + userId + ", email: " + userEmail + "\n";
```
bengaskagoogle: Same thoughts as above.

code/src/main/java/com/onlinecontacttracing/storage/data-storage/PositiveUser.java
```diff
@@ -73,7 +73,7 @@ public boolean userCanMakeMoreDraftsAfterBeingFlagged() {
...
     String hasNumberOfAttemptsLeft = "  number of attempted drafts left: " + (Constants.NUMBER_OF_DRAFTS_ALLOWED - attemptedEmailDrafts) + "\n";
```
bengaskagoogle: I think handling all of this logic in toString() is the wrong place for it. If this is supposed to be something more essential than a toString() dump of the data it probably needs its own methods. At the very least, beware combining this toString() logic which could be called anywhere, with this accounting logic.

code/src/main/webapp/JavaScript/script.js
```diff
@@ -193,8 +194,16 @@ function attachSignin(element, negativeUser) {
   });
 }
 
+function makeRequest(negativeUser, params) {    
+  if (negativeUser) {
+    return new Request('/get-negative-user-info', {method: 'POST', body: params})
+  } else {
+    return new Request('/get-positve-user-info', {method: 'POST', body: params})
```
bengaskagoogle: typo

.gitignore
```diff
@@ -1 +1,2 @@
 code/target/
+code/src/main/webapp/WEB-INF/credentials.json
```
bengaskagoogle: High level comment. 

This is at least 9 different PRs you could have easily split up:

1) Updating the abstract methods signatures and the code that implements them
2) Updating the logic to updateUser and getPayload
3) The GetCalendarFor*Users() stubs
4) GetNegativeUserInfoServlet
5) GetPeopleDataForPositiveUser
6) Changing the error handling case for some reason
7) GetPositiveUserInfoServlet
8) NegativeUser fixes
9) PositiveUser fixes

That along with much of this code not being complete at the time you're asking for review. It's really advantageous to splitting this up because we could approve different simpler things seperately and quickly, leaving the harder chunks of code for discussion.

bengaskagoogle (changes requested):

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: No, I just wanted to show it was handled but I can remove it.

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetCalendarDataForNegativeUser.java
nvergel: There is, all the data for positive users is stored in a list and all the data for negative user is stored individually. This is so we can easily loop through a positive user's data and query a negative user's data

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetCalendarDataForPositiveUser.java
nvergel: Okay, I will change them to nouns

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetNegativeUserInfoServlet.java
nvergel: The data is stored.

Also this is an attempt at modularity, right now we could do it without a runnable but we could only handle one API, if later we wanted to handle two APIs, for example the Drive API, we would have to make two runnables and move all the code into the runnable.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetPositiveUserInfoServlet.java
nvergel: fixed

code/src/main/java/com/onlinecontacttracing/servlets/get-data/GetPositiveUserInfoServlet.java
nvergel: this is meant for when all the data is acquired after implementing the runnables (separate PRs)

code/src/main/java/com/onlinecontacttracing/storage/data-storage/NegativeUser.java
nvergel: okie

code/src/main/java/com/onlinecontacttracing/storage/data-storage/NegativeUser.java
nvergel: yeah it was copy pasted oops

code/src/main/java/com/onlinecontacttracing/storage/data-storage/NegativeUser.java
nvergel: I meant User- ID: _____, email: ______

Like the - refers to ID and email if that makes sense

nvergel: So are you saying to replace it with String.format? Or to keep it in mind for the future?

code/src/main/java/com/onlinecontacttracing/storage/data-storage/PositiveUser.java
nvergel: All this code is from a separate PR, I might just undo the changes in these files because they are distracting from the main point: adding a servlet infrastructure

code/src/main/webapp/JavaScript/script.js
nvergel: fixed

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -35,17 +35,17 @@
...
-  //Update the userId with the newly created credential
-  abstract void updateUser(String userId);
+  // Update the userId with the newly created credential
+  abstract void updateUser(String userId, String email);
 
   private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
   private static final List<String> SCOPES = Collections.singletonList("https://www.googleapis.com/auth/contacts.readonly");
   private static final String CREDENTIALS_FILE_PATH = "WEB-INF/credentials.json";
-  private static final String url = "https://covid-catchers-fixed-gcp.ue.r.appspot.com";
+  private static final String url = "https://8080-49ecfd50-1d05-462f-af38-ebb02e752a59.us-central1.cloudshell.dev";
```
bengaskagoogle: So, I was told this is a debugging one and that the covid-catchers-fixed-gcp.ue.r.appspot.com one is the one you actually want, right?

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
bengaskagoogle: Cool

.gitignore
bengaskagoogle: So, talked with Nico off PR. 

His comments were that it was designed to be large so because "the goal is to set up all the servlets needed for APIs and once the branch is merge we can make PRs that only alter one file (a runnable)."

To which I say, a huge portion of this is unrelated to this still. And we could have also already merged a very large portion of this code because the sticking points are only on a small subset of the files. For instance, the runnable stubs could have already been pushed. The Positive and Negative user fixes are both very minor fixes that could have been handled separately (and are bug fixes anyway that aren't in files that are otherwise modified). 

It honestly makes the entire process faster to have smaller PRs. I think the concern is we have to get it out, so we're doing it in one big PR to get it done. But this really does actually SLOW things down. 

Most of this code could have been merged already if it was split off into its own PR. Everything else is now held up by the discussion on these last parts.

code/src/main/java/com/onlinecontacttracing/storage/data-storage/NegativeUser.java
bengaskagoogle: So you're saying you meant it to be "User- ID:______, (User) Email: _____" where the second User is implied by the "- "?

That's confusing. Just say User-ID and User-Email if that is your intention. 

Also, keep it in mind for the future. I don't care if you do it here, but String.format tends to be way more readable than what you're doing here. It's also slightly faster I think (String concatenation is expensive in Java). Though that's less of an issue.

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: Okay, I removed it

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: yep, I forgot to put it back

.gitignore
nvergel: I guess I was referring mostly to the files related to this PR. But yes you are right, this process is drawn out by all the files that vaguely relate to this branch's goal. I removed those files and will make a new PR for them.

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -61,20 +61,26 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
       log.warning("credentials.json not found");
       response.sendRedirect("/?page=login&error=FileError");
     } catch (GeneralSecurityException e) {
       log.warning("http transport failed, security error");
-      response.sendRedirect("/?page=login&error=TransportError");
-    } catch(Exception e) { //don't expect any other error
-      log.warning("exception occurred");
+      response.sendRedirect("/?page=login&error=GeneralError");
+    } catch(Exception e) { // don't expect any other error
+      log.warning("A exception occurred: " + e.toString());
```
bengaskagoogle: nit: An

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -61,20 +61,26 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
...
-      // Get userId
-      String userId = getUserId(idTokenString, flow, response);
+      // Get payload containing user info
+      Payload payload = getUserId(idTokenString, flow, response);
+
+      // Get userId form payload and retrieve credential
+      String userId = payload.getSubject();
+      String email = payload.getEmail();
+      
       TokenResponse tokenResponse = flow.newTokenRequest(code).setRedirectUri(url+getServletURIName()).execute();
```
bengaskagoogle: nit: add spaces around +

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -61,20 +61,26 @@ public void doGet(HttpServletRequest request, HttpServletResponse response) thro
       String code = request.getParameter("code");
       String idTokenString = request.getParameter("state");
 
-      // Get userId
-      String userId = getUserId(idTokenString, flow, response);
+      // Get payload containing user info
+      Payload payload = getUserId(idTokenString, flow, response);
```
bengaskagoogle: Should this be "getPayload" now?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -0,0 +1,15 @@
+package com.onlinecontacttracing.authentication;
+
+import com.google.api.client.auth.oauth2.Credential;
+
+class CalendarDataForPositiveUser implements Runnable {
```
bengaskagoogle, line 5: So, I still have an underlying question.

Will there be any difference between getting calendar data for a Positive User versus a Negative User? If so, what is it and why is the distinction needed?

If not, merge these.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/PositiveUserInfoServlet.java
```diff
@@ -0,0 +1,61 @@
...
+    Thread contactInfo = new Thread(new CalendarDataForPositiveUser(credential));
+
+    peopleInfo.start();
+    contactInfo.start();
+
+    
+    peopleInfo.join();
+    contactInfo.join();
+
+    // Load PositiveUserContacts from objectify
```
bengaskagoogle: So, this is just a comment without any code. Maybe you intend to put something here? If so, mark it with TODO(...). Else, get rid of this comment.

code/src/main/webapp/JavaScript/script.js
```diff
@@ -179,7 +179,8 @@ function attachSignin(element, negativeUser) {
     localStorage.setItem('idToken', idToken.toString());
     const params = new URLSearchParams()
     params.append('idToken', idToken);
-    fetch(new Request('/authentication-test', {method: 'POST', body: params}))
+
+    fetch(makeRequest(negativeUser, params))
     .then(response => response.text())
     .then(url => { if (url.length < 20) {
```
bengaskagoogle, line 185: Where does this magic value come from?

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: si

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: Yeah, will copy and paste my previous comment since it seems like it got lost in the process:

All the data for positive users is stored in a list and all the data for negative user is stored individually. This is so we can easily loop through a positive user's data and query a negative user's data. Therefore one runnable will put data into a list and store it, the other will store data as it retrieves it.

If this is not clear enough I can elaborate.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/PositiveUserInfoServlet.java
nvergel: Added TODO to every comment that needs it

code/src/main/webapp/JavaScript/script.js
nvergel: I counted the length of the longest string (which was 14) and added a little bit to it just to be safe

nvergel: looking back that may not have been the best solution

I will think of a better approach and put it in a separate PR

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
bengaskagoogle: So, that seems to not answer my question. That seems to be answering why it is Runnable. 

The question is, we have a CalendarDataForPositiveUser and one for CalendarDataForNegativeUser. What's the difference between getting the data for these two different types of users? It seems like they both should be the same Calendar call.

code/src/main/webapp/JavaScript/script.js
bengaskagoogle: Yep.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: Oh no! I copied the wrong message

nvergel: I edited my response with the correct message

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -35,12 +35,12 @@
 @WebServlet("/check-for-api-authorization")
```
gradywardgoogle: We should probably remove this line - since this is an abstract class, it's not instantiable, and would be able to serve a request at this (or any other) URI.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForNegativeUser.java
```diff
@@ -0,0 +1,15 @@
+package com.onlinecontacttracing.authentication;
+
+import com.google.api.client.auth.oauth2.Credential;
+
+class CalendarDataForNegativeUser implements Runnable {
+  Credential credential;
```
gradywardgoogle: nit: private final

code/src/main/java/com/onlinecontacttracing/servlets/get-data/NegativeUserInfoServlet.java
```diff
@@ -0,0 +1,48 @@
...
+import java.util.Optional;
+
+@WebServlet("/get-negative-user-info")
+public class NegativeUserInfoServlet extends CheckForApiAuthorizationServlet {
+
+  String getServletURIName() {
+    return "/get-negative-user-info";
+  }
+  
+  void useCredential(String userId, Credential credential, HttpServletResponse response) throws IOException, InterruptedException {
```
gradywardgoogle, line 23: Here and elsewhere, please use @Override to annotate methods that are overriding a parent declaration.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/NegativeUserInfoServlet.java
```diff
@@ -0,0 +1,48 @@
...
+    Thread contactInfo = new Thread(new CalendarDataForNegativeUser(credential));
+
+    contactInfo.start();
```
gradywardgoogle: .start() .join() can be simplified to .run() (in a single threaded scenario)

code/src/main/java/com/onlinecontacttracing/servlets/get-data/NegativeUserInfoServlet.java
```diff
@@ -0,0 +1,48 @@
...
+  
+    contactInfo.join();
+
+    response.sendRedirect("/?page=notification");
+  }
+
+
+
```
gradywardgoogle: here and elsewhere: 1 line between methods please,

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
gradywardgoogle: If I understand correctly, we actually store data to storage in these runnables, right? If so that could explain the differentiation.

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForNegativeUser.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/get-data/NegativeUserInfoServlet.java
nvergel: do they also go in the run() for runnables?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/NegativeUserInfoServlet.java
nvergel: cool

code/src/main/java/com/onlinecontacttracing/servlets/get-data/NegativeUserInfoServlet.java
nvergel: fixed

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
bengaskagoogle: Wait, so all PositiveUsers are stored in one object, while NegativeUsers are stored one user per object?

bengaskagoogle: If so, that's fine with me. It's difficult to see what's going on here with the skeletons, but if they are actually going to be different, that's that.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/NegativeUserInfoServlet.java
nvergel: nvm I added it there too

bengaskagoogle (approved):

PR #12, opened 2020-07-23
add servlet to delete all negative user data

nvergel:


code/src/main/webapp/index.html
```diff
@@ -81,6 +81,11 @@ <h1> Log In </h1>
       <p> Thank you for signing up. You will be notified if we think
           you may have come in contact with someone who tested positive.
       </p>
+      <p> You can also choose to opt-out and delete all the data we have gathered:
```
gradywardgoogle: maybe also describe "we delete all of your data after 14 days anyway, but you can choose to opt out..."

gradywardgoogle (approved):

PR #18, opened 2020-07-28
Positive user calendar

nvergel:


code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+import com.onlinecontacttracing.storage.PositiveUserContacts;
+import com.google.api.services.calendar.model.EventAttendee;
+import java.util.List;
+import java.util.Collections;
+import java.util.Optional;
+import com.googlecode.objectify.Objectify;
 
 class CalendarDataForPositiveUser implements Runnable {
+
+  private static final String APPLICATION_NAME = "Google Calendar API Java Quickstart";
```
gradywardgoogle: nit: plaease change to reflect our application.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+  private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+
+  private final String userId;
```
gradywardgoogle, line 29: nit: please always order arguments, fields and assignments the same way - a constructor should have all of these in the same order.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+    this.userId = userId;
     this.credential = credential;
+    this.ofy = ofy;
   }
 
   @Override
   public void run() {
-    // TODO Get contacts from people api
+    try {
+      final NetHttpTransport HTTP_TRANSPORT = GoogleNetHttpTransport.newTrustedTransport();
```
gradywardgoogle: why does this variable need to be final?

Also, since this isn't static + final, please make it lowerCamelCaseVariableName.

UPPER_SNAKE is reserved for static final vars.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+
+      // Get user's calendar
+      Calendar service = new Calendar.Builder(HTTP_TRANSPORT, JSON_FACTORY, credential)
+        .setApplicationName(APPLICATION_NAME)
+        .build();
+
+      // Query events between now and the last two weeks
+      long currentTime = System.currentTimeMillis();
+      DateTime now = new DateTime(currentTime);
+      DateTime twoWeeksAgo = new DateTime(currentTime-Constants.CONTACTS_QUERY_TIME);
```
gradywardgoogle: nit - spaces after and before operators.

gradywardgoogle: nit: please rename this to "startOfContactsQueryWindow" or equivalent. We shouldn't have to change this name (or the comment above) if we change the value of CONTACTS_QUERY_TIME

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+      Events events = service.events().list("primary")
```
gradywardgoogle: what does primary here do?

gradywardgoogle: Oh it's the primary calendar? if so please put it in a static constant var.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+        .setTimeMin(twoWeeksAgo)
+        .setTimeMax(now)
+        .execute();
+      List<Event> items = events.getItems();
```
gradywardgoogle: nit: we don't need to name this variable, just put it in the for loop declaration.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+
+      // Initiate objects to store information
```
gradywardgoogle, line 57: What happens if there is already a PostiveUserContacts for this user? I think we'll just overwrite it, yes? Is that the behavior you want?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+      PositiveUserContacts positiveUserContacts = new PositiveUserContacts(userId);
+      
+      // Iterate through events to extract contacts and places
+      for (Event event : items) {
+        Optional<List<EventAttendee>> attendees = Optional.ofNullable(event.getAttendees());
+
```
gradywardgoogle, line 63: nice use of optional! I'd recommend moving the orElse into the previous statement: 

```
List<EventAttendee> attendees = Optional.ofNullable(event.getAttendees()).orElse(Collections.emptyList());

```

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -1,16 +1,74 @@
...
+        for (EventAttendee attendee : attendees.orElse(Collections.emptyList())) {
+          positiveUserContacts.add(attendee.getDisplayName(), attendee.getEmail());
+        }
+      }
+
+      ofy.save().entity(positiveUserContacts).now();
+      
+    } catch (Exception e) {
+      e.printStackTrace();
+      // Not sure what to do here, I tried throwing the exception but then run() isn't overwritten
```
gradywardgoogle: Please log a severe warning!

code/src/main/java/com/onlinecontacttracing/storage/data-storage/Constants.java
```diff
@@ -7,6 +7,7 @@
   final static long SIX_HOURS_IN_SECONDS = 21600L;
   final static long ONE_WEEK_IN_SECONDS = 604800L;
   final static long TWO_WEEKS_IN_SECONDS = 1209600L;
+  final static long TWO_WEEKS_IN_MILLI_SECONDS = 1209600000L;
```
gradywardgoogle: nit: here and above, it's nicer to see this written as a multiplication rather than a long. EX:
```
final static long TWO_WEEKS_IN_MILLI_SECONDS = 2 * 7 * 24 * 60 * 60 * 1000L;
```

code/src/main/java/com/onlinecontacttracing/storage/data-storage/PotentialContact.java
```diff
@@ -21,6 +24,31 @@ public String getEmail() {
     return emailOfPerson;
   }
 
+  /*
+  *  I added equals and hashCode so that we can use a hashSet when adding PotentialContacts
```
gradywardgoogle: Nice, but remove the "I added", instead just saying "equals and hashcode are defined so that... "

gradywardgoogle: Also maybe note here that we treat any two people with the same email address as the same user.

gradywardgoogle (approved):
Nice work, LGTM.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: This is how we have it in all the code

nvergel: It came from the quickstart like this

code/src/main/java/com/onlinecontacttracing/storage/data-storage/Constants.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: my b

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: yep

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: Okay, should we look at other calendars too or just the primary?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: Maybe not, maybe we should not get any data if the user has signed in before?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: Okay

nvergel: Also this could go in the for loop declaration

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: good idea

code/src/main/java/com/onlinecontacttracing/storage/data-storage/PotentialContact.java
nvergel: sure thing

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
gradywardgoogle: OK - well then feel free to fix it everywhere :)

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
gradywardgoogle: Up to you, but I think we want that behavior to be well defined.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
gradywardgoogle: I suggest it not - I try to keep loop declarations on a single line, and usually devoid of logic. Here there's an implicit conditional (the orElse) that probably deserves its own variable.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: I changed it to camel case but I won't make it non final since I am not sure what it does and wouldn't want to mess it up

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: Okay, for now I will just have it replace old data.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: okay

PR #20, opened 2020-07-29
fix error handling in js

nvergel:


bengaskagoogle (approved):

code/src/main/webapp/JavaScript/script.js
```diff
@@ -182,12 +182,7 @@ function attachSignin(element, negativeUser) {
 
     fetch(makeRequest(negativeUser, params))
     .then(response => response.text())
-    .then(url => { if (url.length < 20) {
```
nvergel, line 185: Cynthia pointed out we don't need to check if its an error since the servlet sends a redirect not a string.

PR #25, opened 2020-07-30
add places to positive user

nvergel:


code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -47,6 +62,17 @@ public void run() {
       Calendar service = new Calendar.Builder(httpTransport, JSON_FACTORY, credential)
         .setApplicationName(APPLICATION_NAME)
         .build();
+    
+      // Get apiKey
+      InputStream in = new FileInputStream(new File(CREDENTIALS_FILE_PATH));
+      byte[] apiKeyBytes = new byte[apiKeyLength];
+      in.read(apiKeyBytes);
+      in.close();
```
gradywardgoogle: There are ways to do this without knowing the size of the string: https://stackoverflow.com/questions/309424/how-do-i-read-convert-an-inputstream-into-a-string-in-java

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -47,6 +62,17 @@ public void run() {
       Calendar service = new Calendar.Builder(httpTransport, JSON_FACTORY, credential)
         .setApplicationName(APPLICATION_NAME)
         .build();
+    
+      // Get apiKey
```
gradywardgoogle: Plz remove comment, doesn't add value.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -14,28 +14,43 @@
...
   private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
   private static final String calendarType = "primary";
   static final Logger log = Logger.getLogger(CalendarDataForPositiveUser.class.getName());
+  private static final String CREDENTIALS_FILE_PATH = "WEB-INF/apiKey.txt";
+  private static final int apiKeyLength = 39;
 
   private final Objectify ofy;
   private final String userId;
   private final Credential credential;
   Set<PotentialContact> contacts;
```
gradywardgoogle: I'd prefer these not be instance properties. Please just use return statements to return these back to the calling method. That also allows the helper methods to be static.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -59,18 +85,40 @@ public void run() {
...
+      // Store data or replace old data with newer data.
+      ofy.save().entity(positiveUserPlaces).now();
       
     } catch (Exception e) {
       e.printStackTrace();
       log.warning("An exception occurred: " + e.toString());
     }
   }
+
+  private void getContactsFromEvent(Event event) {
```
gradywardgoogle: please make this a method that returns back the list of contacts, rather than putting them in the object's field. The list of contacts isn't really a propoerty of this object, it's a piece of data created by this object, and used by it. Same in other method below.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
```diff
@@ -59,18 +85,40 @@ public void run() {
...
+    List<EventAttendee> attendees = Optional.ofNullable(event.getAttendees()).orElse(Collections.emptyList());
+
+    for (EventAttendee attendee : attendees) {
```
gradywardgoogle: consider using a stream here instead, using map(...)

code/src/main/java/com/onlinecontacttracing/storage/data-storage/PositiveUserPlaces.java
```diff
@@ -37,4 +37,12 @@ public long getTimeCreatedSeconds() {
     return timeCreatedSeconds;
   }
 
+  public void add(String placeId, String name, long intervalStart, long intervalEnd) {
+    listOfPlaces.add(new Place(placeId, name, intervalStart, intervalEnd));
+  }
+
+  @Override
+  public String toString() {
+    return listOfPlaces.toString();
```
gradywardgoogle: This should probabl also include the user id.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: okay

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: for sure

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: What should the other method return?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: contacts is passed in from the servlet because the servlet needs to use it later on to merge contact sets. How should I make contacts not an instance and still have a pointer to it in the servlet?

code/src/main/java/com/onlinecontacttracing/storage/data-storage/PositiveUserPlaces.java
nvergel: okay

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: I decided to pass in an object to the method but that feels a lot like functional programming and it seems that you are hinting at making the code more OOP so should I move the method getPlacesFromEvent into the PositiveUserPlaces class?

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
gradywardgoogle: Hmmm... perhaps then let's make these changes: 1) still modify the method to return them, and then store that in the run() method to the instance field 2) add a public method called "getContacts" (rather than relying on package visibility

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
gradywardgoogle: That sounds reasonable. What I really want is for our code to not have "side effects". Calling a method that populates a field quietly isn't obvious. Instead, you want to return that data and explicitly store it, or otherwise make that a top-level feature of how the code is working.

code/src/main/java/com/onlinecontacttracing/servlets/get-data/CalendarDataForPositiveUser.java
nvergel: Done

gradywardgoogle (approved):

PR #26, opened 2020-07-31
Page with scopes

nvergel:


code/src/main/webapp/CSS/style.css
```diff
@@ -162,16 +162,16 @@ div {
   transition-delay: 1s
 }
 
-#mission-statement {
+.mission-statement {
```
cynthia-m, line 165: any reason these are classes and not ids anymore? (for all of the ones you changed)

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -57,22 +58,29 @@
...
-      GoogleAuthorizationCodeFlow flow = getFlow(response);
+      // Get parameters from request
       String code = request.getParameter("code");
-      String idTokenString = request.getParameter("state");
+      String stateAsJson = request.getParameter("state");
+
+      // Parse state parameter from Json string to State class
+      Gson gson = new Gson();
+      State state = gson.fromJson(stateAsJson, State.class);
+      List<String> SCOPES = getScopes(state.calendar, state.contacts);
```
cynthia-m: I think this should be names in lowercase because it's not a final var?

cynthia-m (commented):
I think it's good besides 2 tiny things

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: yep

code/src/main/webapp/CSS/style.css
nvergel: id should be unique identifier so I changed repeated ids to classes.

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -38,14 +40,13 @@
 public abstract class CheckForApiAuthorizationServlet extends HttpServlet {
 
   // access API with the created credential
-  abstract void useCredential(String userId, Credential credential, HttpServletResponse response) throws IOException, InterruptedException;
+  abstract void useCredential(State state, Credential credential, HttpServletResponse response) throws IOException, InterruptedException;
```
gradywardgoogle: This is a really good semantic improvement, but let's use a more descriptive term, maybe "PreAuthorizationState"?

gradywardgoogle: of "AuthorizationRoundTripState"

code/src/main/java/com/onlinecontacttracing/servlets/authentication/State.java
```diff
@@ -0,0 +1,17 @@
+package com.onlinecontacttracing.authentication;
+
+/*
+ * Helper class to pass along state information in URL and servlets
+ */
+class State {
```
gradywardgoogle: See comment above about naming. Additionally, can we please use getters rather than package protected access, and can we please use a list of scopes, rather than having booleans for each one?

gradywardgoogle: Actually, perhaps a better solution would be to create an enum called "AuthorizationScopes" (with values today of CALENDAR and CONTACTS). Each takes in as a enum parameter the name of the scope that they correspond to. Then you can have a `Set<AuthorizationScope> authorizedScopes` here, and add an instance method `isAuthorizedForScope(AuthorizationScope scope)` which does a contains check.

code/src/main/webapp/CSS/style.css
gradywardgoogle: +1 good job

code/src/main/webapp/JavaScript/script.js
```diff
@@ -187,8 +187,12 @@ function attachSignin(element, negativeUser) {
     var servlet = "";
     if (negativeUser) {
       servlet = '/get-negative-user-info';
+      params.append('calendar', true);
```
gradywardgoogle, line 190: why don't we give the negative user the same control as the positive user? 

I guess we can't really do anything for them without that control.

code/src/main/webapp/JavaScript/script.js
```diff
@@ -187,8 +187,12 @@ function attachSignin(element, negativeUser) {
...
+      params.append('contacts', false);
     } else {
       servlet = '/get-positive-user-info';
+      params.append('calendar', document.getElementById('calendar').checked);
```
gradywardgoogle, line 194: I think I'd prefer if we attached a list of AuthorizationScopes to the request, rather than individual parameters.
So for example, you'd create a list and then if each of them were checked you would add it to the list. The keywords for this could be the enum names of the AuthorizationScopes in our backend.

code/src/main/java/com/onlinecontacttracing/messaging/HasSpanishTranslation.java
```diff
@@ -0,0 +1,5 @@
+package com.onlinecontacttracing.messaging;
+
+public interface HasSpanishTranslation {
```
gradywardgoogle, line 6: I think this is in a separate PR?

code/src/main/java/com/onlinecontacttracing/servlets/authentication/AuthenticationScope.java
```diff
@@ -0,0 +1,33 @@
...
+  private String scopeName;
+  AuthenticationScope(String scopeName) {
+    this.scopeName = scopeName;
+  }
+
+  public static String getScopeName(AuthenticationScope scope) {
+    for (AuthenticationScope authenticationScope : AuthenticationScope.values()) { 
+      if (authenticationScope.equals(scope)) {
+        return scope.scopeName;
+      }
```
gradywardgoogle: this can actuall all just be replaced by an instance method!

```
public String getScopeName() {
  return this.scopeName;
}
```

code/src/main/java/com/onlinecontacttracing/servlets/authentication/AuthenticationScope.java
```diff
@@ -0,0 +1,33 @@
...
+    return AuthenticationScope.CALENDAR.scopeName;
+  }
+
+  public static AuthenticationScope getScope(String scopeStringName) {
+    for (AuthenticationScope scope : AuthenticationScope.values()) { 
+      if (scope.name().equals(scopeStringName)) {
+        return scope;
+      }
+    }
+    return AuthenticationScope.CALENDAR;
```
gradywardgoogle: Instead of returning something back here, just throw an exception describing that the passed in string is not valid. (IllegalStateException might be a good one)

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
```diff
@@ -134,18 +141,16 @@ public void doPost(HttpServletRequest request, HttpServletResponse response) thr
...
+  // private List<String> getScopes(boolean calendar, boolean contacts) {
+  //   List<String> scopes = new ArrayList<String>();
+  //   if (calendar) {
+  //     scopes.add(CalendarScopes.CALENDAR_READONLY);
+  //   }
+  //   if (contacts){
+  //     scopes.add(PeopleServiceScopes.CONTACTS_READONLY);
+  //   }
+  //   return scopes;
+  // }
```
gradywardgoogle: plz delete

code/src/main/java/com/onlinecontacttracing/servlets/authentication/State.java
```diff
@@ -1,17 +1,39 @@
...
  */
 class State {
   final String idToken;
   String userId;
-  final boolean calendar;
-  final boolean contacts;
+  List<AuthenticationScope> authenticationScopes;
 
-  public State(String idToken, boolean calendar, boolean contacts) {
+  public State(String idToken, List<AuthenticationScope> authenticationScopes) {
```
gradywardgoogle: Can we rename this class AuthorizationRoundTripState?

code/src/main/java/com/onlinecontacttracing/servlets/authentication/State.java
```diff
@@ -1,17 +1,39 @@
...
+  }
+  public static List<AuthenticationScope> getScopes(HashMap<String, Boolean> scopeAuthenticationStatus) {
+    List<AuthenticationScope> scopes = new ArrayList<AuthenticationScope> ();
+    for (String scope : scopeAuthenticationStatus.keySet()) {
+      if (scopeAuthenticationStatus.get(scope)) {
+        scopes.add(AuthenticationScope.getScope(scope));   
+      }
+    }
+    return scopes;
+  }
```
gradywardgoogle: I'm not clear where either of these methods are used?

gradywardgoogle (commented):
Largely LGTM. A few questions about how it works, but otherwise pretty good!

code/src/main/java/com/onlinecontacttracing/messaging/HasSpanishTranslation.java
nvergel: removed

code/src/main/java/com/onlinecontacttracing/servlets/authentication/AuthenticationScope.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/authentication/AuthenticationScope.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/authentication/CheckForApiAuthorizationServlet.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/servlets/authentication/State.java
nvergel: yep

code/src/main/java/com/onlinecontacttracing/servlets/authentication/State.java
nvergel: I also didn't understand what some of the methods were for. I got rid of most of them and tried to maintain the same functionality

code/src/main/java/com/onlinecontacttracing/messaging/HasSpanishTranslation.java
```diff
@@ -1,5 +1,11 @@
 package com.onlinecontacttracing.messaging;
 
+/**
+* Guarantees a message has an English translation
```
gradywardgoogle: Spanish

gradywardgoogle: Same below

gradywardgoogle (approved):
LGTM minus the Spanish/English thing.

code/src/main/java/com/onlinecontacttracing/messaging/HasSpanishTranslation.java
cynthia-m: I see thanks!

PR #32, opened 2020-08-04
Send msg pull

nvergel:


code/src/main/java/com/onlinecontacttracing/messaging/CompiledMessage.java
```diff
@@ -0,0 +1,105 @@
...
+    if (messagesForFrontendDisplay.size() > 1){
+      userMessage = "";
+    }
+    this.messagesForBackendUse = (translatedSystemMessage.concat("\n").concat(userMessage).concat("\n").concat(translatedResourceMessage));
+  }
+
+  /**
+  * Returns the compiled message for backend use.
+  */
+  public String getCompiledBackendMessage() {
```
gradywardgoogle: rather than doing this, let's just make compileMessages(...) return a string! That way we can get rid of this field.

code/src/main/java/com/onlinecontacttracing/servlets/messaging/MessageSendingServlet.java
```diff
@@ -1,36 +1,75 @@
...
-import javax.servlet.annotation.WebServlet;
-import javax.servlet.http.HttpServlet;
 import java.security.GeneralSecurityException;
+import com.onlinecontacttracing.storage.NotificationBatch;
+import static com.googlecode.objectify.ObjectifyService.ofy;
 
-@WebServlet("/message-sender")
-public class MessageSender extends HttpServlet {
 
+@WebServlet("/send-messages")
```
gradywardgoogle: nit - this should go directly on the class, the javadoc should be above it.

code/src/main/java/com/onlinecontacttracing/servlets/messaging/MessageSendingServlet.java
```diff
@@ -1,36 +1,75 @@
...
+      GoogleIdToken idToken = verifier.verify(idTokenString);
+      Payload payload = idToken.getPayload();
+      String userId = payload.getSubject();
+
+      PositiveUser positiveUser = ofy().load().type(PositiveUser.class).id(userId).now();
+      
+      String message = "custom message will be retrieved from params";
+      CompiledMessage compiledMessage = new CompiledMessage(systemMessage, localityResource, message, positiveUser);
+
+      EmailSender.sendEmailsOut("COVID-19 Updates", compiledMessage, messageLanguage);
```
gradywardgoogle: plz make this a constant, and also maybe make it more descriptive?

Example: COVID-19 Potential Exposure Alert

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,116 @@
...
+        contact.markContactedSuccessfully();
+      }
+                  
+    } catch (MessagingException e) {
+        log.warning("error in sending emails out");
+    } catch (GeneralSecurityException e) {
+      log.warning("http transport failed, security error");
+    } catch (Exception e) {
+      log.warning("general exception");
+      e.printStackTrace();
```
gradywardgoogle: you can actually write the error to the log, please do so for all of the exception types.

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,116 @@
...
+import static com.googlecode.objectify.ObjectifyService.ofy;
+
+/**
+* Email sender for sending emails to user's contacts.
+*/
+public class EmailSender {
+
+  private static final JsonFactory JSON_FACTORY = JacksonFactory.getDefaultInstance();
+  private static final List<String> SCOPES = Collections.singletonList(GmailScopes.GMAIL_SEND);
+  private static final String serviceAccountId = "online-contact-tracing@appspot.gserviceaccount.com";
```
gradywardgoogle: nit: UPPER_SNAKE_CASE same with next one.

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,116 @@
...
+import java.io.InputStream;
+import java.security.GeneralSecurityException;
+import java.io.ByteArrayOutputStream;
+import java.util.Properties;
+import javax.mail.Session;
+import javax.mail.internet.InternetAddress;
+import static com.googlecode.objectify.ObjectifyService.ofy;
+
+/**
+* Email sender for sending emails to user's contacts.
```
gradywardgoogle: plz include more detail - who will get the emails, who will they be sent by, etc.

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,116 @@
...
+        .build();
+
+      Gmail service = new Gmail.Builder(HTTP_TRANSPORT, JSON_FACTORY, serviceAccountCredential)
+                  .setApplicationName(APPLICATION_NAME)
+                  .build();
+                  
+      compiledMessage.compileMessages(messageLanguage);
+      String emailBody = compiledMessage.getCompiledBackendMessage();
+
+      NotificationBatch notificationInfo = ofy().load().type(NotificationBatch.class).id(compiledMessage.getUserId()).now();
```
gradywardgoogle, line 72: can a notification batch include emails sent in multiple languages? If so, we'll need to remove the message language above, and compile it for each user individually.

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,116 @@
...
+      log.warning("general exception");
+      e.printStackTrace();
+    }
+  }
+
+  /**
+  * Sends out messages to a specific contact.
+  */
+  private static void sendMessage(String to, String emailSubject, String emailContent, Gmail service) throws MessagingException, IOException {
+    System.out.println(to);
```
gradywardgoogle: plz remove.

code/src/main/java/com/onlinecontacttracing/servlets/messaging/MessageSendingServlet.java
gradywardgoogle: Also, actually this doesn't work because the title should change based on the language specified. Can you make EmailSubject a class that implements HasEnglishTranslation/HasSpanishTranslation?

gradywardgoogle (commented):
Great work across the board. Needs some tweaks, but very close.

code/src/main/java/com/onlinecontacttracing/servlets/messaging/MessageSendingServlet.java
nvergel: done

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
nvergel: No, right now it is set up to send the same message to every person

cynthia-m: ^^^ we will have another branch to take care of this

code/src/main/java/com/onlinecontacttracing/messaging/CompiledMessage.java
gradywardgoogle: I still see this as a bad design - can we please fix?

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
gradywardgoogle: ah

code/src/main/java/com/onlinecontacttracing/messaging/EmailSubject.java
```diff
@@ -0,0 +1,30 @@
+package com.onlinecontacttracing.messaging;
+
+/*
+* System's opening message.
```
gradywardgoogle: Plz give a better javadoc comment?

Also I think this class should implement two interfaces :) and that means the methods should have overrides. I also think it should probably be an Enum (similar to the way we did other message types.

code/src/main/java/com/onlinecontacttracing/messaging/EmailSender.java
```diff
@@ -0,0 +1,114 @@
...
+  private static final String EMAIL_TO_SEND_WITH = "cccoders@onlinecontacttracing.com";
+  private static final String CREDENTIALS_FILE_PATH = "online-contact-tracing-f798898872f4.p12";
+  private static final String TOKENS_DIRECTORY_PATH = "tokens";
+  private static final String APPLICATION_NAME = "Online Contact Tracing";
+  static final Logger log = Logger.getLogger(EmailSender.class.getName());
+
+  /**
+  * Sends emails with a email subject and a message to a list of contacts.
+  */
+  public static void sendEmailsOut(String emailSubject, CompiledMessage compiledMessage, String messageLanguage) {
```
gradywardgoogle: Please pass in the EmailSubject Object rather than passing in the string it returns.

gradywardgoogle (approved):
Approving after last comment is resolved.

PR #35, opened 2020-08-05
Delete positive user data

nvergel:


code/src/main/java/com/onlinecontacttracing/servlets/delete-data/DeleteDataAfterApproval.java
```diff
@@ -0,0 +1,52 @@
...
+        .setAudience(Collections.singletonList(CLIENT_ID))
+        .build();
+
+      // Get payload with userId
+      String idTokenString = request.getParameter("idToken");
+      GoogleIdToken idToken = verifier.verify(idTokenString);
+      Payload payload = idToken.getPayload();
+      String userId = payload.getSubject();
+
+      ofy().delete().type(PositiveUserContacts.class).id(userId).now();
```
bengaskagoogle: This is a very nitty thing, but I'd move this into a function called delete, so all the specifics are contained there, and you just pass in userId. Very much encapsulates that logic.

bengaskagoogle (approved):

PR #36, opened 2020-08-05
fix bug where user is displayed

nvergel:


code/src/main/java/com/onlinecontacttracing/servlets/get-data/PositiveUserInfoServlet.java
```diff
@@ -54,11 +53,13 @@ void useCredential(AuthorizationRoundTripState state, Credential credential, Htt
     calendarInfo.join();
 
     PositiveUserContacts fullContacts = ofy().load().type(PositiveUserContacts.class).id(state.userId).now();
-    if(fullContacts.getListOfContacts().isEmpty()) {
+    if(fullContacts == null) {
```
bengaskagoogle, line 56: Do you want to check for both the null case and the empty contact list case?

bengaskagoogle (approved):

code/src/main/java/com/onlinecontacttracing/servlets/get-data/PositiveUserInfoServlet.java
nvergel: It doesn't matter if it is empty, the set will still add data.

nvergel: Also in the runnable that populates that list, it checks if the list is empty before saving so they should be mutually exclusive.

bengaskagoogle: Sounds good to me. Feel free to resolve.

nvergel: meant to say mutually inclusive*

PR #40, opened 2020-08-06
add redirects for pages

nvergel:


code/src/main/java/com/onlinecontacttracing/servlets/redirects/NegativeLoginServlet.java
```diff
@@ -0,0 +1,16 @@
+package com.onlinecontacttracing.servlets.redirects;
+
+import javax.servlet.annotation.WebServlet;
+import javax.servlet.http.HttpServlet;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.io.IOException;
+
+@WebServlet("/negative-login")
```
gradywardgoogle, line 9: I don't really get what this does? What is the page= notation?

nvergel: the javascript for the landing page reads the params in the url to check which page to load

gradywardgoogle: oh right, cool. Thanks!

